PR #8, opened 2016-05-19
add libzip dependency to README

prt2121:
re Build Failure https://github.com/rahulmutt/ghcvm/issues/7


rahulmutt:
Let's get that C preprocessor part working as well and then we'll update the README in one go.

prt2121:
looks like you guys are trying to get rid of libzip. i'm closing this now

PR #18, opened 2016-09-06
Need to cast closures to StgMutVar

puffnfresh:


puffnfresh:
Had code which wasn't passing the verifier without this.

Issue #19, opened 2016-09-06
Integrate compilation of java, class, jar files


rahulmutt:
```
ghcvm --make SomeClass.class SomeOtherClass.java SomeJar.jar SomePure.hs
```
Should convert all the given files to class files accordingly and properly put them in the resulting jar. Currently, it's partially implemented (should work in batch mode `-c`) but needs to be tweaked for `--make` mode. In particular, we need to make sure we set the `-classpath` flag of `javac` properly, taking into account references to the module closures compiled in the same batch.

Implementation Sketch:
Separate out .java files and compile them only after all other compilation units and include all the compiled .jars as dependencies.

PR #26, opened 2016-09-07
Implement c_isatty

puffnfresh:


puffnfresh:
In the runtime until #19 is sorted.

PR #29, opened 2016-09-09
Make "print" work without the infinite loop

puffnfresh:


libraries/base/GHC/IO/Encoding/Iconv.hs
```diff
@@ -58,12 +59,12 @@ iconv_trace s
 
 {-# NOINLINE localeEncodingName #-}
```
zjhmale: @puffnfresh one question, why we need `NOINLINE` pragma here? i mean, we don't use something like `inlinePerformIO`, so what exactly the reason that we should put it here?

rahulmutt: @zjhmale The original implementation in GHC's base had NOINLINE because it used unsafePerformIO. We will eventually use that implementation once the equivalent of c_localeEncoding is implemented.

puffnfresh: And we do use that, `c_localeEncoding` was implemented sometime yesterday :cat2:

zjhmale: @rahulmutt  thx for the answers! i just missed `unsafePerformIO` :sweat_smile:

PR #64, opened 2016-09-25
Initial travis CI integration

psibi:
Creates a oracle jdk 8 environment along with stack and right now it just runs `verify.sh`. We can expand on this further. Also I have added `set -e` on `verify.sh`.


rahulmutt:
Great! Thanks a lot.

PR #83, opened 2016-10-17
Script to test few selected packages

psibi:
This is a hacky script which does the job now. I will improve on this further tomorrow. We need a proper way of storing the ported packages in some configuration file now. I will open the relevant proposals as issues tomorrow.

### DIFF
--- a/tests/packages/Test.hs
+++ b/tests/packages/Test.hs
@@ -21,9 +21,9 @@ patchedPackages
     :: [String]
 patchedPackages =
     [ "array"
+    , "bytestring"
     , "base-orphans"
     , "binary"
-    , "bytestring"
     , "containers"
     , "filepath"
     , "primitive"
@@ -38,7 +38,6 @@ unPatchedPackages =
     [ "Adaptive"
     , "agum"
     , "array-utils"
-    , "base-orphans"
     , "base-prelude"
     , "basic-lens"]
 
--- a/tests/packages/Test.hs
+++ b/tests/packages/Test.hs
@@ -55,6 +55,8 @@ buildPackage pkg = do
 
 main :: IO ()
 main = do
+  let vmUpdateCmd = "cabalvm update"
+  shell vmUpdateCmd ""
   cabalvmPkgs <- packagesFilePath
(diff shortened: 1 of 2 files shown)


rahulmutt:
Great start to the script. I agree that there should be files in ghcvm-hackage that indicate the packages that should be tested. 

Before merging this, I'll need you to do a couple things:
1. Create all the files you need in ghcvm-hackage and submit a PR to that repository.
2. Update the Test.hs script to read from the files instead.
    Note that ghcvm-hackage is essentially cloned to `~/.cabalvm/patches/` after running `cabalvm update` so you can look there for the files you created in step 1.
3. Make the Travis build pass. If any of the packages fail, let me know and I'll update the patches to fix the build.

tests/packages/Test.hs
```diff
@@ -0,0 +1,62 @@
...
+    parseJSON _ = empty
+
+parsePackagesFile :: FilePath -> IO (Maybe Packages)
+parsePackagesFile fname = do
+  contents <- BS.readFile fname
+  let packages = decode contents
+  return packages
+
+packagesFilePath :: IO FilePath
+packagesFilePath = (<> "/.cabalvm/packages.json") <$> getHomeDirectory
```
rahulmutt: Comments:

1. This is not portable since you're using '/' directly - use `</>` instead.

2. In Windows, there won't be a `.cabalvm` because the `.` is *nix-specific convention for application data directories. Prefer `getAppUserDataDirectory` instead.

3. The path is `~/.cabalvm/patches/packages.json` - sorry that's a bit confusing. I'm open to better naming for the directory where `ghcvm-hackage` should be cloned.

All three are fixed below:
```haskell
(\x -> x </> "patches" </> "packages.json") <$> getAppUserDataDirectory "cabalvm"  
```
You can write it in a style that suits you, but the semantics should match.

psibi: Thanks for the feedback, I will fix them by tonight likely.

psibi: Note that I'm using `</>` from the package `filepath`. Let me know if you had something different in mind.

PR #105, opened 2016-10-24
Grammer and Formatting fix

psibi:


psibi:
@rahulmutt I think this can be merged. Is something blocking here ?

PR #129, opened 2016-11-20
Add OS X Travis Integration

psibi:
Right now, OS X just runs till `verify.sh`. Only the Linux container, builds the entire package set.


rahulmutt:
That'll do for now. Verify.sh catches most of the major bugs.

PR #134, opened 2016-11-27
Add show instance for JString

psibi:


libraries/base/ETA/JString.hs
```diff
@@ -28,12 +28,17 @@ import GHC.Real
...
+import GHC.Show (Show(..))
+import GHC.Pack (unpackCString)
 
 -- TODO: Add rules to simplify
 -- fromString (unpackCString# "Hello world!"#) :: JString = JString "Hello world"#
 instance IsString JString where
   fromString = mkJString
 
+instance Show JString where
+    show = unpackCString
```
puffnfresh: Do we want this to be consistent with `Show String`, where the quotes are put around the result?

psibi: We can make it consistent by using `show . unpackCString`. But I don't have any strong opinions on it. I think making it equivalent to `Show String` would be nice as people may be relying on this behaviour:

```
show string == show jstring
```

rahulmutt: For now, add quotes around the result and as we write more libraries with Java bindings, we'll get an understanding of whether it needs to be modified.

PR #141, opened 2016-12-05
Use touchFile on Windows as GHCs touchy is not available

alexbiehl:
As `eta` doesn't supply a `touchy` equivalent as GHC does. This path sidesteps the issue by doing the `touch` *in* `eta`. This is a very unclean solution. If the user supplies a custom path to `touch` or `touchy` with `-pgm_T` the path gets ignored on Windows.

### DIFF
--- a/compiler/ETA/Main/SysTools.hs
+++ b/compiler/ETA/Main/SysTools.hs
@@ -88,6 +88,7 @@ import qualified System.Posix.Internals
 #else /* Must be Win32 */
 import Foreign
 import Foreign.C.String
+import System.PosixCompat.Files
 #endif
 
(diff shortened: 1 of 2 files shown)


rahulmutt:
What a coincidence, you just resolved #142! I don't mind if you do away with`SysTools.touch` completely - could you take a stab at that?

alexbiehl:
@rahulmutt You mean completly removing `SysTools.touch` and changing the implementation of `touchObjectFile`?

rahulmutt:
Yes exactly.

alexbiehl:
Yes, I can do that immediately

Issue #142, opened 2016-12-05
Add portable implementation of `touchObjectFile`


rahulmutt:
In `ETA.Main.DriverPipeline`, `touchObjectFile` calls upon `SysTools.touch` to execute the system `touch` command. To avoid dependencies on external tools to keep the compiler cross-platform (especially Windows-happy), it would be best if we had a Haskell implementation (using C FFI is also fair game, but it would be preferred if it could be avoided).

A potential implementation to look at can be found in the `turtle` [package](https://hackage.haskell.org/package/turtle-1.3.0/docs/src/Turtle-Prelude.html#touch).

PR #143, opened 2016-12-05
Use portable `touchFile` implementation in `touchObjectFile`

alexbiehl:
This is the patch discussed in https://github.com/typelead/eta/pull/141. This fixes https://github.com/typelead/eta/issues/142.


rahulmutt:
Looks good. I'll merge it as soon as Travis gives me thumbs up.

rahulmutt:
The build gave this error:
`dist/build/GHC/IO.jar-boot: touchFile: does not exist (No such file or directory)`

You need to handle the case where it doesn't exist. You should create an empty file in that case.

alexbiehl:
Travis is green now!

PR #145, opened 2016-12-06
Use unix-compat-0.4.3.0

alexbiehl:


alexbiehl:
(I think stack needs another minute to add it to all-cabal-hashes repo)

alexbiehl:
There is a build error on appveyor related to the patch

PR #156, opened 2017-01-04
Resolves #15 - Named inner classes for RtsFuns

### LINKED ISSUE #15: Named inner classes
Currently, anonymous inner classes are used to declare RtsFuns. Instead, we should used named inner classes so that debugging becomes clearer. An example is `eta.runtime.apply.Apply` which contains a lot of them which are frequently used. It's easier to decipher what `Apply$ApPFast` is rather than `Apply$8`.

alexander-at-github:
Use named inner classes instead of ananymous classes to declare RtsFuns.

### DIFF
--- a/rts/src/eta/runtime/apply/Apply.java
+++ b/rts/src/eta/runtime/apply/Apply.java
@@ -6,208 +6,229 @@ import eta.runtime.stg.RtsFun;
 import eta.runtime.stg.AbstractArgumentStack;
 
 public class Apply {
-    public static RtsFun ap_0_fast = new RtsFun() {
-            @Override
-            public void enter(StgContext context) {
-                StgClosure fun = context.R(1);
-                fun.evaluate(context);
-            }
-        };
+    public static RtsFun ap_0_fast = new Ap0Fast();
+    public static RtsFun ap_v_fast = new ApVFast();
+    public static RtsFun ap_n_fast = new ApNFast();
+    public static RtsFun ap_l_fast = new ApLFast();
+    public static RtsFun ap_f_fast = new ApFFast();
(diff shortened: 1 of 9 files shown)


rts/src/eta/runtime/apply/Apply.java
```diff
@@ -6,208 +6,229 @@
...
-                AbstractArgumentStack stack = (AbstractArgumentStack) context.O(1);
-                context.merge(stack);
-                StgClosure fun = pap.fun;
-                fun.enter(context);
-            }
-        };
+    public static RtsFun ap_0_fast = new Ap0Fast();
+    public static RtsFun ap_v_fast = new ApVFast();
+    public static RtsFun ap_n_fast = new ApNFast();
+    public static RtsFun ap_l_fast = new Ap1Fast();
```
rahulmutt: Should be `ApLFast`, not `Ap1Fast`

rts/src/eta/runtime/apply/Apply.java
```diff
@@ -6,208 +6,229 @@
...
+    public static RtsFun ap_f_fast = new ApFFast();
+    public static RtsFun ap_d_fast = new ApDFast();
+    public static RtsFun ap_o_fast = new ApOFast();
+    public static RtsFun ap_p_fast = new ApPFast();
+    public static RtsFun ap_pv_fast = new ApPvFast();
```
rahulmutt: Should be `ApPVFast`, make sure you capitalise all the letters after `Ap` and before `Fast`. Do the same for all the remaining `Ap*Fast`

alexander-at-github:
I amended the requested changes into my commit.

PR #162, opened 2017-01-12
[WIP] etai

psibi:


shake/Build.hs
```diff
@@ -111,6 +111,7 @@ fixGhcPrimConf = do
 buildLibrary :: Bool -> String -> [String] -> Action ()
 buildLibrary debug lib deps = do
   let dir = library lib
+      externalDepLibs = ["ghc-boot"]
```
rahulmutt: Can you put this at the top level since it's constant? Easier to change later on.

libraries/base/GHC/IO/Exception.hs
```diff
@@ -231,7 +232,7 @@ data ExitCode
                 -- The exact interpretation of the code is
                 -- operating-system dependent.  In particular, some values
                 -- may be prohibited (e.g. 0 on a POSIX-compliant system).
-  deriving (Eq, Ord, Read, Show, Typeable)
+  deriving (Eq, Ord, Read, Show, Generic, Typeable)
```
rahulmutt: This should mostly be harmless, but it may create some inconsistencies when compiling Hackage packages. You can keep this as it is, but can you add a warning here that this may lead to some problems later on?

libraries/ghc-boot/changelog.md
```diff
@@ -0,0 +1,5 @@
+## 8.0.1  *May 2016*
+
+  * Bundled with GHC 8.0.1
+
+  * Initial version
```
rahulmutt: Probably better to omit this file altogether. It may create some confusion. Or replace the contents instead with the commit hash of the GHC repo as a version for `ghc-boot` so we can keep track as we integrate GHC 8 changes going forward.

rahulmutt:
This PR has not been active for more than over a month. Closing. You can send a new PR if you get time to take another stab at this.

PR #168, opened 2017-01-14
Add brew installation instructions

cema-sp:
I've created a homebrew formula for easier installation of **eta** on MacOS.

Please, review the formula too: https://github.com/cema-sp/homebrew-tap/blob/master/Formula/eta.rb


rahulmutt:
The formula looks good, I'll try it out and merge if it works.

PR #179, opened 2017-01-16
Fix typo in FFI example

vic:


rahulmutt:
That is not actually a typo :) You can't have variables starting with an uppercase letter in Eta because that is reserved for types.

vic:
Oh my bad then, closing this then. *you never seen this*.

Thanks for eta btw !

rahulmutt:
No issues! Feel free to PR anything you feel is off and we can get it clarified. I'd rather have that than miss out on some typo.

PR #181, opened 2017-01-16
Fix minor typo in the Getting Started page

beickhoff:
For the ```Counter``` class example, ```public int set``` should be ```public void set```.  This matches the imported type signature:

    foreign import java unsafe set :: Int -> Java Counter ()


rahulmutt:
Thanks for the catch!

PR #182, opened 2017-01-16
Added WSL method to the end of source installation route

MrDWilson:


rahulmutt:
Looking good. Just above that there's a warning that the windows installation doesn't work. Can you modify that warning and reference the sections you added?

MrDWilson:
I've amended to warning. Let me know if there is anything else you want changing.

PR #193, opened 2017-01-21
windows support

jneira:


rahulmutt:
Looks great @jneira! Now I just need you to add info to the docs:

jneira:
Hi, i tested the build in win xp and 10 and it seems ok

PR #197, opened 2017-01-22
Filter invalid files like *.dsa, *.sf in the jar

psibi:
Since we copy the existing jar file to a location, we also need to
inspect it and clean it up if it's already present. We also do the
proper cleaning up in the new files before they are written to the
archive.

### DIFF
--- a/compiler/ETA/Utils/JAR.hs
+++ b/compiler/ETA/Utils/JAR.hs
@@ -1,3 +1,5 @@
+{-#LANGUAGE OverloadedStrings#-}
+
 {-|
 Module      : ETA.Utils.JAR
 Description : Small utility functions for creating Jars from class files.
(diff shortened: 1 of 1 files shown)


compiler/ETA/Utils/JAR.hs
```diff
@@ -282,9 +287,19 @@ mergeClassesAndJars jarLocation compress fileAndContents jarSelectors = do
   p <- makeAbsoluteFilePath jarLocation
   copyFile copy p
   withArchive p $ do
+    existingEntries <- getEntries
+    let invalidEntries = filterWithKey (\k _ -> invalidEntrySelector k)  existingEntries
+    mapM_ deleteEntry (keys invalidEntries)
```
rahulmutt: This line can have unintended consequences for the user. Was it necessary to fix the problem you faced?

psibi: @rahulmutt Oh yeah. This is the actual fix. The problem is that the copied file initially has those files. If you see the link `copyFile copy p` will copy the jar file to a new location. And then we are acting on that jar file with the function `withArchive`.

rahulmutt: My bad, you're right! I misread the code.

PR #199, opened 2017-01-22
Maven docs improve

psibi:
@rahulmutt I also slightly changed the wording in one case and added some more information from the gitter chat we had.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -1295,6 +1295,12 @@ In this example, we'll be binding to the `Unirest <http://unirest.io/java>`_ lib
 Setup a project, just like :ref:`setting-up-first-project` with the following
 changes:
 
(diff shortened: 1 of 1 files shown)


docs/source/getting-started.rst
```diff
@@ -1282,6 +1284,9 @@ Maven repositories:
 \*\* ``[owner]`` should be the owner of the  repository and ``[repo]`` should
      be the name of the Bintray repository.
 
+By default, ``epm`` tries to resolve your dependencies against the
+above repositories. So, you don't need to specify a repository unless required.
```
rahulmutt: Should be '..resolve your dependencies against ``central``'. The other repositories specified in the table don't get used unless you specify them.

psibi: @rahulmutt Ah right, I understand it now. :-) Updated.

PR #218, opened 2017-02-04
Adds missing array exports array(To/From)List

aktowns:
Noticed that the `arrayToList` and `arrayFromList` in the documentation http://eta-lang.org/docs/html/getting-started.html#working-with-arrays were not exported (unsure if on purpose and I'm missing something obvious), this fixed the problem for me


rahulmutt:
Thanks for the catch! I must've accidentally removed them when writing the `JavaConverter` instance.

Issue #221, opened 2017-02-06
Builder (ByteString/Binary etc) return Prelude.undefined


aktowns:
Not sure if this is related to lack of C FFI? Playing around it appears

```haskell
-- Data-ByteString-Internal
mallocByteString :: Int -> IO (ForeignPtr a)
````
Always returns a null pointer, they also 'show' as abnormally large
`0x0000000000000000000000000000000000000000000000000000000000000000`.  digging through the source for `Data,ByteString.Builder` it looks like
new buffers are created this way, somehow eventually they end up as `undefined`. A small example
that replicates it is:

```haskell
import qualified Data.ByteString.Builder as B
main = print $ B.toLazyByteString $ B.byteString "a"
```
which throws
```
Prelude.undefined
java.lang.Exception: Stack trace
	at java.lang.Thread.dumpStack(Thread.java:1333)
	at eta.runtime.RtsMessages.errorBelch(RtsMessages.java:32)
	at eta.base.Utils.errorBelch(Utils.java:154)
	at base.ghc.conc.Sync$satzusEKLU.enter(Unknown Source)
	at eta.runtime.apply.Apply$PAPApply.enter(Apply.java:231)
	at eta.runtime.apply.StgPAP.apply(StgPAP.java:46)
	at eta.runtime.apply.ApV.stackEnter(ApV.java:12)
	at eta.runtime.stg.StackFrame.enter(StackFrame.java:43)
	at eta.runtime.stg.StgContext.checkForStackFrames(StgContext.java:79)
	at base.ghc.Foreign$zdwa1.enter(Unknown Source)
	at base.ghc.Foreign$zdwa4zusD25E.enter(Unknown Source)
	at base.ghc.Foreign$satzusD265.enter(Unknown Source)
	at eta.runtime.apply.Apply$PAPApply.enter(Apply.java:231)
	at eta.runtime.apply.StgPAP.apply(StgPAP.java:46)
	at eta.runtime.apply.ApV.stackEnter(ApV.java:12)
	at eta.runtime.stg.StackFrame.enter(StackFrame.java:43)
	at eta.runtime.stg.StackFrame.enter(StackFrame.java:26)
	at eta.runtime.stg.StgContext.checkForStackFrames(StgContext.java:79)
	at base.ghc.IO$bracket1.enter(Unknown Source)
	at base.ghc.Foreign$charIsRepresentable3.enter(Unknown Source)
[...]

PR #222, opened 2017-02-07
Implements memmove, memcpy and memset for Foreign.Marshal

aktowns:
Fixes the `Prelude.undefined` panic in #221, 

Haven't dealt with ByteBuffers before but its almost
a verbatim copy of the implementation in the bytestring patch https://github.com/typelead/eta-hackage/blob/master/patches/bytestring-0.10.8.1.patch 

Since we already dup the src and copy its my understanding its safe to use the same method for both memcpy and memmove there's just no performance gain with memcpy

### DIFF
--- a/libraries/base/java-utils/Utils.java
+++ b/libraries/base/java-utils/Utils.java
@@ -18,6 +18,7 @@ import eta.runtime.Rts;
 import eta.runtime.RtsFlags;
 import static eta.runtime.Rts.ExitCode;
 import eta.runtime.RtsMessages;
+import eta.runtime.io.MemoryManager;
 
 import java.lang.management.ManagementFactory;
 import com.sun.management.OperatingSystemMXBean;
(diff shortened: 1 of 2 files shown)


libraries/base/java-utils/Utils.java
```diff
@@ -123,6 +123,23 @@ public static String byteBufferToStr(ByteBuffer buffer)
         return new String(bytes, "UTF-8");
     }
 
+    public static ByteBuffer c_memcpy(ByteBuffer dest, ByteBuffer src, int size) {
+        src = src.duplicate();
+        ByteBuffer dest2 = dest.duplicate();
+        src.limit(src.position() + size);
```
rahulmutt: In `eta.runtime.io.MemoryManager`, there's a function called `bufSetLimit` that does the equivalent of what you're doing here, but it's more robust to changes in the MemoryManager.

Change `src.limit(src.position() + size)` to `bufSetLimit(src, size)`.

libraries/base/java-utils/Utils.java
```diff
@@ -123,6 +123,23 @@ public static String byteBufferToStr(ByteBuffer buffer)
...
+        dest2.put(src);
+        return dest;
```
rahulmutt: Return `dest2` instead of `dest`. Always return new objects to keep all the operations pure. That's why we do a lot of `duplicate()`'s when dealing with `ByteBuffer`'s. This may not be a problem, but it's good to stay safe.

aktowns:
Thanks for the review, I've updated with those changes, and made sure i'm duping and returning the dups for the ByteBuffers. 

I didn't realise `eta.runtime.io.MemoryManager` exists thats awesome, for things like `malloc` java wrappers elsewhere I guess I should be using `eta.runtime.io.MemoryManager.allocateBuffer` so it gets thrown into the addressMap?

rahulmutt:
Yeah exactly. Freeing off-heap memory/reorganizing it to avoid defragmentation haven't been implemented yet. For now free can be treated as a no op.

libraries/base/java-utils/Utils.java
```diff
@@ -123,6 +124,34 @@ public static String byteBufferToStr(ByteBuffer buffer)
         return new String(bytes, "UTF-8");
     }
 
+    public static ByteBuffer c_memcpy(ByteBuffer dest, ByteBuffer src, int size) {
+        ByteBuffer src2 = src.duplicate();
+        ByteBuffer dest2 = dest.duplicate();
+        MemoryManager.bufSetLimit(src2, size);
+        dest2.put(src2);
+        return dest2;
```
rahulmutt: Apologize - my last claim about returning `dest2` was wrong! I thought about it some more and realised that the returned buffer *must* point to the start of `dest` while `dest2` will point to after the newly written data. So return `dest`. Do the same for `memmove` as well and it should be good.

aktowns:
All good, updated PR reverted it on both memmove/memcpy, cheers!

PR #223, opened 2017-02-07
Adds missing toUnsignedBigInteger(I)

aktowns:
The second thing in #221 
Also changes the padding on Ptr's Show to match ghc's on 64bit platforms (8 bytes)

### DIFF
--- a/libraries/base/GHC/Ptr.hs
+++ b/libraries/base/GHC/Ptr.hs
@@ -31,6 +31,8 @@ import GHC.Num
 import GHC.List ( length, replicate )
 import Numeric          ( showHex )
 
+#include "MachDeps.h"
+
 ------------------------------------------------------------------------
 -- Data pointers.
 
@@ -165,7 +167,7 @@ instance Show (Ptr a) where
      where
         -- want 0s prefixed to pad it out to a fixed length.
        pad_out ls =
-          '0':'x':(replicate (2*8 - length ls) '0') ++ ls ++ rs
+         '0':'x':(replicate (2*SIZEOF_HSPTR - length ls) '0') ++ ls ++ rs
 
 instance Show (FunPtr a) where
    showsPrec p = showsPrec p . castFunPtrToPtr
--- a/libraries/base/GHC/Ptr.hs
+++ b/libraries/base/GHC/Ptr.hs
@@ -165,7 +165,7 @@ instance Show (Ptr a) where
      where
         -- want 0s prefixed to pad it out to a fixed length.
        pad_out ls =
-          '0':'x':(replicate (2*32 - length ls) '0') ++ ls ++ rs
+          '0':'x':(replicate (2*8 - length ls) '0') ++ ls ++ rs
 
 instance Show (FunPtr a) where
    showsPrec p = showsPrec p . castFunPtrToPtr
--- a/libraries/integer/java/Utils.java
+++ b/libraries/integer/java/Utils.java
@@ -46,7 +46,7 @@ public class Utils {
      * Return a BigInteger equal to the unsigned value of the
      * argument.
      */
-    private static BigInteger toUnsignedBigInteger(long i) {
+    public static BigInteger toUnsignedBigInteger(long i) {
         if (i >= 0L)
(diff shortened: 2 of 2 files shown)


libraries/integer/java/Utils.java
```diff
@@ -59,6 +59,10 @@ private static BigInteger toUnsignedBigInteger(long i) {
         }
     }
 
+    public static BigInteger toUnsignedBigInteger(int i) {
+      return Utils.toUnsignedBigInteger(Long.valueOf(i));
```
rahulmutt: Very subtle bug here: if you pass in negative integers, you'll get negative BigIntegers which is not the intended behaviour. Use `toUnsignedLong` instead of `Long.valueOf`.

libraries/base/GHC/Ptr.hs
```diff
@@ -165,7 +165,7 @@ instance Show (Ptr a) where
      where
         -- want 0s prefixed to pad it out to a fixed length.
        pad_out ls =
-          '0':'x':(replicate (2*32 - length ls) '0') ++ ls ++ rs
+          '0':'x':(replicate (2*8 - length ls) '0') ++ ls ++ rs
```
rahulmutt: This looks ok, just want to hear the rationale behind the change. Why 32 -> 8?

aktowns: Ahh because its a hex representation of the bytes of the address `Ptr` so 8 bytes on 64bit, with 2 0's for each byte (this is at least how haskell on 64bit machines do it) on 32bit this wont make sense as it should be smaller (4*2) (looks like GHC itself uses `SIZEOF_HSPTR`, but i don't think we have access to that?)

libraries/integer/java/Utils.java
aktowns: Ahh i originally put that, but was getting a weird result, since `-1` ends up padded then converted it ends up as
```
-1 (input)
0x00000000ffffffff (output)
```
where as passing the long directly in to get converted i get
```
-1 (input)
0xffffffffffffffff (output)
```
I guess the actual problem is a 32bit integer is being used to represent a pointer on a 64bit platform so there is data loss in the conversion? I'll switch it over to use `toUnsignedLong` when i get home

rahulmutt:
We actually do have access to SIZEOF_HSPTR inside of MachDeps.h (was a recent change). In Eta, the effective word size is 32-bit regardless of platform. There are no real pointers on the JVM, so we're just simulating it. This decision was made as a result of a limitation in ByteBuffers only taking Java ints as indexing arguments.

aktowns:
Ahh sorry my bad I mistakenly assumed the word size was 64, that makes a lot more sense. 

I've updated the PR to use `toUnsignedLong` and changed the padding to use `SIZEOF_HSPTR` so it matches up with upstream ghc's source. Output looks good from my small test case

```haskell
-- $ cat src/Main.hs 
{-# LANGUAGE OverloadedStrings, GHCForeignImportPrim, MagicHash, BangPatterns, ScopedTypeVariables #-}
module Main where

import Foreign.Ptr
import Foreign.Marshal(alloca)
import GHC.Word
import GHC.Prim
import GHC.Integer
import GHC.Exts
import Data.Int

dummyAllocator :: IO ()
dummyAllocator = do
  _ <- alloca (\(ptr :: Ptr Int64) -> print ptr)
  return ()

main :: IO ()
main = do
  dummyAllocator
  dummyAllocator
  dummyAllocator
  print $ I# (addr2Int# nullAddr#)
  print nullPtr

{- 
$ runhaskell src/Main.hs
0x00000002001ca390
0x00000002001ca3a8
0x00000002001ca3c0
0
0x0000000000000000       
$ epm run
Preprocessing executable 'eta-bytestring-builder' for
eta-bytestring-builder-0.1.0.0...
Running eta-bytestring-builder...
0x00000000
0x00005fa8
0x00005fb0
-1
0xffffffff
-}
```

rahulmutt:
Looks good now! If you're not sure about anything, you can feel free to ask because a lot of things aren't documented. Best place for quick answers is Gitter or Slack.

PR #229, opened 2017-02-17
Use POSIX string equality operator

amir:
shebang is looking for sh but is using bash's `==` operator to compare strings.

According to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html correct operator is `=`.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -3,7 +3,7 @@
 _pull=$(git pull)
 _subs=$(git submodule update --recursive)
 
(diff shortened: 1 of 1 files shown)


amir:
To further clarify here's the output of running `update.sh` before this fix on a Debian with `sh` being linked to `dash`:

```
λ eta git:(master) ./update.sh
./update.sh: 6: [: Already up-to-date.: unexpected operator

```

rahulmutt:
Thanks for the catch!

Issue #106, opened 2016-10-25
Windows support


alexbiehl:
I am trying to bootstrap `eta` on Windows and `stack exec eta-build install` fails with 

```
Configuring ghc-prim-0.4.0.0...
# epm (for install)
Building ghc-prim-0.4.0.0...
Preprocessing library ghc-prim-0.4.0.0...
eta.exe: could not execute: C:\Users\abiehl\AppData\Roaming\eta/../mingw/bin/gcc.exe
Error when running Shake build system:
* install
user error (Development.Shake.cmd, system command failed
Command: epm build
Current directory: C:\Users\abiehl\git\eta\libraries\ghc-prim
Exit code: 1
Stderr:
eta.exe: could not execute: C:\Users\abiehl\AppData\Roaming\eta/../mingw/bin/gcc.exe
)
```

Somehow it thinks a `gcc` is at `C:\Users\abiehl\AppData\Roaming\eta/../mingw/bin/gcc.exe` which is of course not right! (a good `gcc` is on path through `stack exec`). 

Trying to manually build `ghc-prim` with 
`stack exec -- epm build --eta-options=" -v4 -pgmP C:\Users\abiehl\AppData\Local\Programs\stack\x86_64-windows\ghc-7.10.3\mingw\bin\gcc"`

```
Building ghc-prim-0.4.0.0...
Preprocessing library ghc-prim-0.4.0.0...
The ETA programming language compiler, Version 0.0.1 booted by GHC version 7.10.3
wired-in package ghc-prim not found.
wired-in package integer not found.
wired-in package base not found.
wired-in package rts mapped to rts-0.1.0.0-dd60ce55e5e6e7d613c6ae1fe388a75d
wired-in package template-haskell not found.
wired-in package ghc not found.
wired-in package dph-seq not found.
wired-in package dph-par not found.
E   rts-0.1.0.0-dd60ce55e5e6e7d613c6ae1fe388a75d
Hsc static flags:
wired-in package ghc-prim not found.
wired-in package integer not found.
wired-in package base not found.
wired-in package rts mapped to rts-0.1.0.0-dd60ce55e5e6e7d613c6ae1fe388a75d
wired-in package template-haskell not found.
wired-in package ghc not found.
wired-in package dph-seq not found.
wired-in package dph-par not found.
*** Chasing dependencies:
Chasing modules from: *GHC.CString,*GHC.Classes,*GHC.Magic,*GHC.Tuple,*GHC.Types,*GHC.JArray
[...]

PR #231, opened 2017-02-21
Windows support iconv

jneira:
Changes to windows.support (#106):
- Translating debug.sh to debug.cmd
- Implement errno handling in java
- Translating iconv c calls in java ones

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,6 @@ install:
 
 script:
 # Build the package, its tests, and its docs and run the tests
-- cd tests/verify/
-- ./verify.sh
 - |
   if [ `uname` != "Darwin" ]
   then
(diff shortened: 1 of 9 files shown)


jneira:
one comment about performance: the java decode step uses an intermediate output buffer as bigger as the input one so the program will use more memory than the ghc version

rahulmutt:
Great work @jneira! Please rebase one more time because I've updated the testing scripts for Appveyor and Travis.

jneira:
@rahulmutt  i've done the rebase (with some strange problems) and it seems to be correct

rahulmutt:
Thanks! It looks like the AppVeyor build printed "Hello World!". I'll just wait for it to compile the rest of the packages and see if it encounters the patches problem like you did.

rahulmutt:
The AppVeyor build failed, but that seems to be an orthogonal problem. I'll merge.

Issue #31, opened 2016-09-10
Add an error upon defining "identical" haskell functions, types, and data constructors


rahulmutt:
When the user defines two top-level functions that are equivalent modulo case, they will currently get a NoClassDef error when running the program. We should stop this prematurely after renaming, informing the user that he shouldn't do this. For now we'll mention it clearly in the docs.

rahulmutt:
Copied from Gitter to preserve in a more persistent place:

As Eta is derived from GHC, a lot of the staged architecture is similar (up to the code generator at least). In the frontend of the compiler, the first couple stages are parsing and renaming. Renaming is the stage where all the names are fully qualified with the original source of the definition.
For this issue in particular, you’d need to add a pass in the renamer where you loop through all the names that are defined in the module and check whether two of the names share the same case, and if so, print a nice error message.

The actual working functions of the renamer are defined in `ETA.Rename.*`
In particular, check out `rnSrcDecls` in `ETA.Rename.RnSource`

You may need to use some monadic functions of the `RnM` monad. Checkout `ETA.TypeCheck.TcRnMonad` and `ETA.TypeCheck.TcRnTypes`.

And `ETA.HsSyn.*` for the definitions of all the types used in the full AST.

The typechecker and renamer share the same monad and in fact `RnM` is just a type synonym for `TcM`.

rahulmutt:
Closing since the PR relating to this has been merged.

PR #245, opened 2017-03-25
Print error message and exit on similar names.

pparkkin:
Find identifiers with names differing only in case. If found, print error message and exit.

See issue https://github.com/typelead/eta/issues/31.

### DIFF
--- a/compiler/ETA/BasicTypes/Avail.hs
+++ b/compiler/ETA/BasicTypes/Avail.hs
@@ -70,10 +70,16 @@ availsToNameEnv avails = foldr add emptyNameEnv avails
 findSames :: [AvailInfo] -> [[Name]]
 findSames as = filter (\l -> length l > 1) sames
   where
-    sames = M.elems mm ++ M.elems am
-    (mm, am) = foldr addTo (M.empty, M.empty) as
-    addTo a (mm, am) = (addName n mm, foldr (\n m -> if isDataConName n then addName n m else m) am ans)
-      where (n:ans) = availNames a
+    sames = M.elems mm ++ M.elems dm ++ M.elems vm
+    (mm, dm, vm) = foldr addTo (M.empty, M.empty, M.empty) as
+    addTo a (mm, dm, vm) =
+        ( addName n mm
+        -- Separate maps for data constructors and variables
+        -- Don't want to mix them up
+        , foldr addName dm (filter isDataConName ns)
+        , foldr addName vm (filter isVarName ns)
+        )
+      where (n:ns) = availNames a
     addName n m =
       case M.lookup l m of
         Just ns -> M.insert l (n : ns) m
--- a/compiler/ETA/BasicTypes/Avail.hs
+++ b/compiler/ETA/BasicTypes/Avail.hs
@@ -68,16 +68,18 @@ availsToNameEnv avails = foldr add emptyNameEnv avails
                                 (zip (availNames avail) (repeat avail))
 
 findSames :: [AvailInfo] -> [[Name]]
-findSames as = filter (\l -> length l > 1) (M.elems sames)
+findSames as = filter (\l -> length l > 1) sames
   where
-    sames = foldr addTo M.empty as
-    addTo a m =
+    sames = M.elems mm ++ M.elems am
+    (mm, am) = foldr addTo (M.empty, M.empty) as
+    addTo a (mm, am) = (addName n mm, foldr addName am ans)
+      where (n:ans) = availNames a
+    addName n m =
       case M.lookup l m of
         Just ns -> M.insert l (n : ns) m
         Nothing -> M.insert l [n] m
-      where n = availName a
-            l = toLower $ occNameString $ nameOccName n
-            toLower = map C.toLower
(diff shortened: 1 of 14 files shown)


rahulmutt:
Great start @pparkkin! Some feedback:

High Priority:
- In the very beginning of `rnSrcDecls`, there's a call to `getLocalNonValBinders` (located in `ETA.Rename.RnNames`) where it checks for duplicate names. It makes sense to integrate your pass in that function somewhere so that it can fail earlier and do less work.
- Try to use the existing error throwing mechanism in the `TcM` monad. Checkout the functions `addErr`, `failWith`, `addErrAt`, `addErrs` in `ETA.TypeCheck.TcRnMonad`.

Optional:
- The runtime complexity of the function that detects similar names is `O(n^2)`, can you make it a fold with a `Map` that stores the lower case names as it runs the pass to bring it down to say `O(nlog(n))`? This is not too much of a priority, but if it sounds interesting to work on you can give it a try.

pparkkin:
Thanks for the review!

I don't think `getLocalNonValBinders` handles the function names at all. To me it looks like they're part of `val_decls`, which is handled in `rnTopBindsLHS` (really `rnValBindsLHS`). That's where I moved the checking to.

I also use `failWith` now. Right now the error message shows line/column as `1:1`. I'll try to find out whether I can get the correct line number.

I'll also do the optimization you mention when I get some more time.

pparkkin:
Implemented the requested optimization, and changed it to output line numbers.

Looks something like this now.
```
Main.hs:8:1:
    Multiple declarations with names differing only in case.
    Declared at: Main.hs:5:1
                 Main.hs:8:1
```

rahulmutt:
Awesome work @pparkkin! The build failed because of the following error:
```haskell
lib/Data/Time/Format/Parse.hs:179:1:
    Multiple declarations with names differing only in case.
    Declared at: lib/Data/Time/Format/Parse.hs:135:1
                         lib/Data/Time/Format/Parse.hs:179:1
```
Indeed, `Data.Time.Format.Parse` uses `readSTime` and `readsTime` in the same file! In practice, this would've caused a runtime bug when parsing time formats. Thanks to your changes, we can now detect them beforehand!

So what I'll do is update the time patches and rerun the build until it works. From your side, everything seems to be fine.

rahulmutt:
I just noticed that the following warnings:
```
/home/travis/build/typelead/eta/compiler/ETA/Rename/RnBinds.hs:50:1: Warning:
    The import of ‘sortBy, intercalate, groupBy’
    from module ‘Data.List’ is redundant
/home/travis/build/typelead/eta/compiler/ETA/Rename/RnBinds.hs:305:22: Warning:
    Defined but not used: ‘name’
```
Can you fix those please?

pparkkin:
Done.

I was looking for a way to get the warnings to show. I see them now. Seems like I just missed them in the `stack` output before.

rahulmutt:
Great. By the way, do your changes work on data constructors and type constructors as well?

pparkkin:
Looks like no. I'll try to get that fixed as soon as I can.

Also, I found out it doesn't only check toplevel definitions, but it will also error out on something like this.

```haskell
module Main where

main :: IO ()
main = do
  let
    hW = "helloWorld"
    hw = "helloworld"
  print hW
  print hw
```

I'll get that sorted out too.

rahulmutt:
No problem. Since you've done the core algorithm, I don't think it'll be too much trouble. While there isn't a proper test suite yet, can you create a single isolated file (or many files) that covers that various cases? I can also add in a couple that I have in mind. You can add the files inside of `tests/basic`.

pparkkin:
- Added some test files
- Support types and constructors
- Only check top level definitions

rahulmutt:
Looks like you fixed that last build failure, great. Now it's failing on some hackage packages. Will upload the patches and see if I can merge this by tonight. Are there any more pending items from your side?

pparkkin:
I'm not sure if this one is a bug on my side, or something with the library being compiled. I was going to take a look at it tonight.

```
Data/DList.hs:163:1:
    Multiple declarations with names differing only in case.
    Declared at: Data/DList.hs:141:1
                 Data/DList.hs:163:1
```

The code in question uses `pattern`. I'm not sure how I should handle it, and I was going to test to see how it works.

```haskell
pattern Cons x xs <- (toList -> x:xs)
...
cons x xs   = DL ((x:) . unDL xs)
```

If you think it's not a bug with my changes, I'm okay with merging right now.

pparkkin:
Pushed a change for handling pattern synonyms. Hopefully it will take care of the issue in the build.

rahulmutt:
You can run tests locally with `stack test eta` to quickly check whether it worked or not. Travis rebuilds `eta` from scratch so it'll take longer. You've already gotten a built copy with your changes so it'll be faster for you.

pparkkin:
Travis passed. I looked through the AppVeyor log, but couldn't see anything related to my changes.

I think we're good.

rahulmutt:
Great, I think that's a good sign that it's working. Merging.

PR #251, opened 2017-04-01
Fix typo installating -> installing

sgronblo:


Jyothsnasrinivas:
Thanks for the catch!

Issue #268, opened 2017-04-07
NPE for wai-servlet sample application


rahulmutt:
Here's the stack trace I get:
```
java.lang.NullPointerException
	bytestring.data.bytestring.Internal$zdwunpackAppendCharsLazzy.enter(Unknown Source)
	wai.network.wai.Internal$satzusCZZL.thunkEnter(Unknown Source)
	eta.runtime.thunk.StgInd.enter(StgInd.java:12)
	eta.runtime.stg.StgClosure.evaluate(StgClosure.java:23)
	base.ghc.Base$zpzp.enter(Unknown Source)
	base.ghc.Base$satzusAQ8.thunkEnter(Unknown Source)
	eta.runtime.thunk.StgInd.enter(StgInd.java:12)
	eta.runtime.stg.StgClosure.evaluate(StgClosure.java:23)
	base.ghc.Base$zpzp.enter(Unknown Source)
	base.ghc.Base$satzusAQ8.thunkEnter(Unknown Source)
	eta.runtime.thunk.StgInd.enter(StgInd.java:12)
	eta.runtime.stg.StgClosure.evaluate(StgClosure.java:23)
	main.Main$zdfstableZZC0ZZCmainZZCMainZZCservShowReq2.enter(Unknown Source)
	main.Main$satzusPL3.enter(Unknown Source)
	eta.runtime.apply.Apply$PAPApply.enter(Apply.java:231)
	eta.runtime.apply.StgPAP.apply(StgPAP.java:172)
	eta.runtime.apply.Apply$ApPVFast.enter(Apply.java:105)
	blaze_builder.blaze.bytestring.Builder$zdwa2zus1EEU.enter(Unknown Source)
	blaze_builder.blaze.bytestring.Builder$satzus1EJ0.enter(Unknown Source)
	eta.runtime.apply.StgFun.apply(StgFun.java:19)
	eta.runtime.apply.ApV.stackEnter(ApV.java:12)
	eta.runtime.stg.StackFrame.enter(StackFrame.java:43)
	eta.runtime.stg.StgContext.checkForStackFrames(StgContext.java:79)
	base.ghc.IO$unsafeDupablePerformIO.enter(Unknown Source)
	blaze_builder.blaze.bytestring.Builder$zdwtoLazzyByteStringWith.enter(Unknown Source)
	main.Main$zdwa.enter(Unknown Source)
	main.Main$zdfstableZZC0ZZCmainZZCMainZZCservShowReq1.enter(Unknown Source)
[...]

PR #314, opened 2017-04-19
Change translation of __NULL literal in bytecode generation phase

jneira:
To fix the bug #268 
I've tested manually with this [code](https://github.com/jneira/eta-test/blob/master/app/Main.hs):
```haskell
main = do
  putStrLn $ "nullPtr: " ++ show nullPtr
  let (fptr,offset,length) = BSInt.toForeignPtr B.empty
  withForeignPtr fptr $ \ ptr -> do
    trace ptr offset length
    putStrLn $ "ptr: " ++ show ptr
    putStrLn $ "ptr == nullPtr " ++ (show $ nullPtr == ptr)
  putStrLn $ "ByteString.empty: " ++  show B.empty
  putStrLn $ "Wai request: " ++ show defaultRequest
```
with this output:
```
Ptr: java.nio.HeapByteBuffer[pos=4 lim=4 cap=4]
Ofset: 0
Length: 0
nullPtr: 0xffffffff
ptr: 0xffffffff
ptr == nullPtr True
ByteString.empty: ""
Wai request: Request {requestMethod = "GET", httpVersion = HTTP/1.0, rawPathInfo = "", rawQueryString = "", requestHeaders = [], isSecure = False, remoteHost = 0.0.0.0:0, pathInfo = [], queryString = [], requestBody = <IO ByteString>, vault = <Vault>, requestBodyLength = KnownLength 0, requestHeaderHost = Nothing, requestHeaderRange = Nothing}
```

### DIFF
--- a/compiler/ETA/CodeGen/Utils.hs
+++ b/compiler/ETA/CodeGen/Utils.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
 module ETA.CodeGen.Utils where
 
 import ETA.Main.DynFlags
@@ -27,8 +26,6 @@ cgLit (MachFloat r)         = (jfloat, fconst $ fromRational r)
 cgLit (MachDouble r)        = (jdouble, dconst $ fromRational r)
 -- TODO: Remove this literal variant?
 cgLit MachNullAddr          = (jobject, nullAddr)
-  where nullAddr = getstatic $ mkFieldRef memoryManager "nullAddress"
-                               byteBufferType
 cgLit (MachStr s)           = (jstring, sconst $ decodeUtf8 s)
 -- TODO: Implement MachLabel
 cgLit MachLabel {}          = error "cgLit: MachLabel"
--- a/compiler/ETA/CodeGen/Utils.hs
+++ b/compiler/ETA/CodeGen/Utils.hs
@@ -10,6 +10,7 @@ import Codec.JVM
 import Data.Char (ord)
 import Control.Arrow(first)
 import ETA.CodeGen.Name
+import ETA.CodeGen.Rts
 import ETA.Debug
 -- import Data.Text (Text)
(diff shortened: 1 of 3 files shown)


compiler/ETA/CodeGen/Utils.hs
```diff
@@ -24,7 +25,9 @@ cgLit (MachWord64 i)        = (jlong, lconst $ fromIntegral i)
 cgLit (MachFloat r)         = (jfloat, fconst $ fromRational r)
 cgLit (MachDouble r)        = (jdouble, dconst $ fromRational r)
 -- TODO: Remove this literal variant?
-cgLit MachNullAddr          = (jobject, aconst_null jobject)
+cgLit MachNullAddr          = (jobject, nullAddr)
+  where nullAddr = getstatic $ mkFieldRef "eta/runtime/io/MemoryManager"
+                   "nullAddress" (obj "java/nio/ByteBuffer")
```
rahulmutt: Couple changes:
- `(obj "java/nio/ByteBuffer")` -> `byteBufferType` (defined in `ETA.CodeGen.Rts`)
- `"eta/runtime/io/MemoryManager"` -> `memoryManager` (defined in `ETA.CodeGen.Rts`)

rahulmutt:
This change will break the handling of `Maybe` arguments in the FFI (probably not used by anyone so probably no one will see the breakage?), because it does a (horrible) unsafeCoerce of `nullAddr#` in order to get `null`. I'll go ahead and file an issue tracking that breakage and we'll eventually get around to fixing it. I'll merge this pull request once it passed the build for the time being.

compiler/ETA/CodeGen/Utils.hs
jneira: mmm ok, otho you think would be better to use a static method to get the constant?

jneira:
finally i've thought that the better place to set the `nullAddr :: Code` is in Rts itself

PR #319, opened 2017-04-24
Make closure initialisation lazy

### LINKED ISSUE #202: Lazy initialization of Eta modules
As shown in #200, strict class loading can wreak havoc on startup times. We should find a way to lazily initialize the closures in an Eta module. Most likely, the solution will involve reflection since that's the only way we can refer to a class without the JVM loading it.

puffnfresh:
Cleans up the code a bit and should make performance a bit better. Need to
measure the impact.

Fixes #202


puffnfresh:
Do we also need to update this?

https://github.com/typelead/eta/blob/15445e21f3ee4e5583b856ec977030ca2a9b048c/compiler/ETA/DeSugar/DsForeign.hs#L555-L557

puffnfresh:
Makes a huge difference to Proguard. Given this program:

```haskell
add x y = x + y + 1
main = print $ add 1 2
```

Before this commit, Proguard took about 5 minutes and gave:

```
19M     RunSimple.jar
7.5M    RunSimpleOpt.jar
```

Now Proguard runs for only 17 seconds and gives:

```
20M     RunSimple.jar
512K    RunSimpleOpt.jar
```

puffnfresh:
We should do some more rigorous performance measurements but this initially looks pretty great for such a tiny program!

Before:

```
$ time java -jar RunSimple.jar 
4

real    0m0.930s
user    0m2.343s
sys     0m0.088s

$ time java -jar RunSimpleOpt.jar 
4

real    0m0.681s
user    0m1.732s
sys     0m0.074s
```

After:

```
$ time java -jar RunSimple.jar 
4

real    0m0.319s
user    0m0.516s
sys     0m0.033s

$ time java -jar RunSimpleOpt.jar 
4

real    0m0.164s
user    0m0.245s
sys     0m0.019s
```

rahulmutt:
@puffnfresh Will merge this as soon as the build is fixed. Also, yes to your question about changing the field reference in foreign exports.

compiler/ETA/CodeGen/Constr.hs
```diff
@@ -41,18 +41,24 @@ cgTopRhsCon dflags id dataCon args = (cgIdInfo, genCode)
...
+                  new dataFt
+                , dup dataFt
+                , fold loads
+                , invokespecial (mkMethodRef dataClass "<init>" fields void)
+                , putstatic field
+                ]
+          defineMethod . mkMethodDef' [Public, Static] qClName (mkMethodDesc [] (Just closureType)) $ fold
+            [
+              getstatic field
+            , markStackMap
```
rahulmutt: Is this needed? No jump happens to this location.

compiler/ETA/CodeGen/Constr.hs
```diff
@@ -41,18 +41,24 @@ cgTopRhsCon dflags id dataCon args = (cgIdInfo, genCode)
...
+          let field = mkFieldRef modClass qClName closureType
+              loadCodes =
+                [
+                  new dataFt
+                , dup dataFt
+                , fold loads
+                , invokespecial (mkMethodRef dataClass "<init>" fields void)
+                , putstatic field
+                ]
+          defineMethod . mkMethodDef' [Public, Static] qClName (mkMethodDesc [] (Just closureType)) $ fold
```
rahulmutt: Why use `mkMethodDef'` vs `mkMethodDef`?

compiler/ETA/CodeGen/Main.hs
```diff
@@ -89,50 +89,53 @@ cgTopRhsClosure dflags recflag id _binderInfo updateFlag args body
...
+               defineMethod $ mkMethodDef' [Public, Static] qClName (mkMethodDesc [] (Just closureType)) $ fold
                  [
-                   new indStaticType,
-                   dup indStaticType,
-                   loadCode,
-                   invokespecial $ mkMethodRef stgIndStatic "<init>"
-                     [closureType] void,
-                   putstatic field
+                   getstatic field
+                 , markStackMap
```
rahulmutt: Same comment as above.

compiler/ETA/CodeGen/Constr.hs
puffnfresh: It kept failing verification for `ifnonnull` until I put in a `markStackMap` - I just chucked it in a random place and the verification problem disappeared. Should it go somewhere else or is the problem actually codec-jvm's handling of `ifnonnull`?

compiler/ETA/CodeGen/Constr.hs
puffnfresh: `mkMethodDef` requires a class name right?

compiler/ETA/CodeGen/Constr.hs
puffnfresh: I guess I should mark where the jump actually is.

```
Exception in thread "main" java.lang.VerifyError: Expecting a stackmap frame at branch target 19
Exception Details:
  Location:
    main/Main.DZCmain_closure()Leta/runtime/stg/StgClosure; @3: ifnonnull
  Reason:
    Expected stackmap frame at this location.
  Bytecode:
    0x0000000: b200 11c7 0010 bb00 0559 b700 16b3 0011
    0x0000010: a700 03b2 0011 b0                      

        at eta.main.main(Unknown Source)
```

But why does this fix the above?

compiler/ETA/CodeGen/Constr.hs
rahulmutt: Yeah, that should be the class name of the class that contains the static method. `mkMethodDef` appends some local variables initialisation code in front of the code you provide and in this case since there are no local variables or arguments, so it doesn't seem to make a difference. I'll file an issue on `codec-jvm` to not export `mkMethodDef'`. Can you plug in the value returned from `getModClass` into that argument for consistency?

compiler/ETA/CodeGen/Constr.hs
rahulmutt: I wonder if the `mkMethodDef'` is causing the problem. Can you try the previous change and let me know if that fixes it without using `markStackMap`?

puffnfresh:
@rahulmutt ah, `mkMethodDef` has an `initCtrlFlow`. I missed that.

I updated the PR to use `mkMethodDef` and removed the `markStackMap` calls.

Issue #322, opened 2017-04-25
Convert stack-based calls to static method calls


rahulmutt:
The program
```haskell
add x y = x + y
```
will generate
```java
package main;

public class Main {
  private StgClosure add_closure;
  public static class add {
    public void enter(StgContext context) {
      StgClosure x = context.R(2);
      StgClosure y = context.R(3);
      // Non-trivial part of entry code    
    }
  }
}
```

The problem here is that this forces you to pass arguments through the stack (`context`)  which is *very* inefficient when your program enters a handful of closures over and over again (very common in most programs)! It would be much more efficient if we could directly pass the arguments to a static method. We can, if we change the code generation very slightly:
```java
package main;

public class Main {
  public StgClosure add_closure = new add();
  public static class add {
    public void enter(StgContext context) {
      StgClosure x = context.R(2);
      StgClosure y = context.R(3);
      add_entry(context, x, y);
    }
  }
  public static void add(StgContext context, StgClosure x, StgClosure y) {
    // Non-trivial part of entry code
  }
}
```
So now, when the program calls the *fully applied* `add` function, the static function `add` will be called directly, bypassing the `add_closure` field access. Moreover, if `add` is applied partially (i.e. used in an `StgPAP`) `add_closure` will be used instead.

rahulmutt:
Saturated (and even over-saturated calls) are all generated in `ETA.CodeGen.Layout.directCall` and `directCall'`. `cgIdApp` is the top-level function that handles all types of function calls.

rahulmutt:
We will not pursue this further since it is can actually *hurt* performance in the general case, with only a few special (and uncommon) cases benefitting from this.

PR #334, opened 2017-04-29
Generate "body" methods and directly call them with arguments

puffnfresh:
This fixes the example from #322. I think there's probably more places where this optimisation can be used.

### DIFF
--- a/compiler/ETA/CodeGen/Bind.hs
+++ b/compiler/ETA/CodeGen/Bind.hs
@@ -21,7 +21,6 @@ import ETA.Util
 import Data.Maybe (catMaybes)
 import ETA.Main.Constants
 import Codec.JVM
-import Codec.JVM.Opcode (nop)
 import Control.Monad (forM, foldM)
 import Data.Text (unpack)
 import Data.Foldable (fold)
--- a/compiler/ETA/CodeGen/Bind.hs
+++ b/compiler/ETA/CodeGen/Bind.hs
@@ -88,8 +88,7 @@ closureCodeBody _ id lfInfo args arity body fvs binderIsFV recIds = do
       --       if there will be a recursive call later. This will
       --       have a significant effect on the size of the resulting
       --       class files.
-      emit $ op nop
-          <> markStackMap
+      emit $ markStackMap
           <> startLabel label
       withSelfLoop (id, label, cgLocs) $ cgExpr body
     return ()
--- a/compiler/ETA/CodeGen/Bind.hs
+++ b/compiler/ETA/CodeGen/Bind.hs
@@ -21,6 +21,7 @@ import ETA.Util
 import Data.Maybe (catMaybes)
 import ETA.Main.Constants
 import Codec.JVM
+import Codec.JVM.Opcode (nop)
 import Control.Monad (forM, foldM)
 import Data.Text (unpack)
 import Data.Foldable (fold)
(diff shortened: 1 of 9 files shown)


compiler/ETA/CodeGen/Layout.hs
```diff
@@ -142,28 +143,52 @@ slowCall fun args = do
...
 
-directCall :: Bool -> Code -> RepArity -> [StgArg] -> CodeGen ()
-directCall slow entryCode arity args = do
+directCall :: Bool -> CgLoc -> RepArity -> [StgArg] -> CodeGen ()
+directCall slow cgLoc@(LocStatic _ modClass clName) arity args = do
+  argFtCodes <- getRepFtCodes args
+  emit $ directStaticCall slow staticClosure modClass' arity argFtCodes
+  where staticClosure =
+             loadLoc cgLoc
+          <> pop closureType
```
puffnfresh: I originally had this as:

```haskell
gconv closureType (obj (innerClas modClass clName))
```

Casts are meant to be fast on the JVM. Faster than this field access? Not sure which it should be.

default.nix
```diff
@@ -21,7 +21,7 @@ let
       tasty-ant-xml = haskell.lib.doJailbreak super.tasty-ant-xml;
       binary = haskell.lib.dontCheck self.binary_0_8_4_1;
 
-      codec-jvm = self.callPackage ./utils/nix/codec-jvm.nix { };
+      codec-jvm = haskell.lib.dontHaddock (self.callPackage ./utils/nix/codec-jvm.nix { });
```
rahulmutt: I fixed this so you can probably revert this. You'll have to bump codec-jvm again.

compiler/ETA/CodeGen/Bind.hs
```diff
@@ -67,23 +67,29 @@ closureCodeBody _ id lfInfo args arity body fvs binderIsFV recIds = do
...
       n <- peekNextLocal
       let (argLocs, code, n') = mkCallEntry n args
-          (_ , cgLocs) = unzip argLocs
-      emit code
+      emit $ gload thisFt 0 <> code
       setNextLocal n'
       bindArgs argLocs
+      mapM_ bindFV fvLocs
+      emit $ invokevirtual (mkMethodRef thisClass "body" bodyType void)
+    _ <- withMethod [Public] "body" bodyType void $ do
```
rahulmutt: Shouldn't this be a static method defined in `modClass`? I think this explains why you're not seeing any speedups :) It's an instance method defined in an inner class of `modClass`. The name of the static method should be identical to name of the closure. And `closureCodeBody` is used for both top-level and non top-level functions so you need to check that there are no free variables before applying the optimisation.

rahulmutt:
Can you rebase this PR with `master`? The updated script that should trigger eta-benchmarks is there.

compiler/ETA/CodeGen/Constr.hs
```diff
@@ -41,8 +41,8 @@ cgTopRhsCon dflags id dataCon args = (cgIdInfo, genCode)
         -- typeFt = obj (tyConClass dflags (dataConTyCon dataCon))
         genCode = do
           loads <- mapM getArgLoadCode . getNonVoids $ zip maybeFields args
-          defineField $ mkFieldDef [Private, Static] qClName closureType
-          let field = mkFieldRef modClass qClName closureType
+          defineField $ mkFieldDef [Public, Static] qClName dataFt
```
rahulmutt: We typically avoid typing fields or method arguments with anything other than `closureType` since we want to be able to rewrite closures at runtime. The JIT will take care of remembering the types for us so I don't think it'll be a huge difference. Note that I did do this actively in the first iteration of the codegen and I had to revert to using `closureType` because there are corner cases where making it too typed is a problem. 

Let me make that concrete with an example. Suppose we have

```
data Test = Test 1 2 3
test :: Test
test = Test 1 2 3
```

And suppose we later redefine it to an `Int` (in cases like an interpreter)

```
test :: Int
test = 1
```

If it was an `StgClosure`, this is very easy to change since it's just a matter of resetting the field. If it was more specific, it would be impossible to place it in the field since Java doesn't allow unsafe coercing.

compiler/ETA/CodeGen/Types.hs
```diff
@@ -157,7 +158,7 @@ mkCgIdInfoWithLoc id lfInfo cgLoc =
            , cgLocation = cgLoc }
 
 mkStaticLoc :: DynFlags -> Id -> LambdaFormInfo -> CgLoc
-mkStaticLoc dflags id _ = LocStatic closureType modClass clName
+mkStaticLoc dflags id _ = LocStatic (obj (qualifiedName modClass clName)) modClass clName
```
rahulmutt: `locFt` used throughout the codegen will extract this field and some corner cases may break the verifier because of this. Can you explain why you made this change? I don't see you using this change anywhere.

rahulmutt:
According to @puffnfresh, performance gains don't seem to be as expected. Need to revisit this at a later time.

rahulmutt:
I had another go at this in the `going-stackless` branch and only a few benchmarks truly benefitted from this optimisation while the others had significant slow downs. Essentially, this optimisation works great when most of your program has function calls which are known at compile-time (which is not the common case in a lazy language).

Another thing to think about is whether the programs in nofib are true representatives of code you'd find it production, which I'm not sure is the case. 

The plan of action to take this forward is add some benchmarks to `eta-benchmarks` with common subsets of "real-world" code and see how this optimisation affects those benchmarks.

Closing for now until we find an extremely common piece of code that could benefit from this optimisation.

Issue #180, opened 2017-01-16
Bug with Unicode strings: ArrayIndexOutOfBoundsException


flyfish30:
Source Code:
```haskell
main :: IO ()
main = mapM_ putStrLn $ sierpinskiS 2

infixl 5 ===
(===) = (  )

infixl 5 |||
(|||) = zipWith (  )

center n = map ((spaces   ) . (   spaces))
    where spaces = replicate (2 ^ (n - 1)) ' '

sierpinskiS :: Int -> [[Char]]
sierpinskiS 0 = ["△ "]
sierpinskiS n = center n s
                ===
                (s ||| s)
    where s = sierpinskiS (n - 1)
```

Output:
```shell
Compiling

rahulmutt:
Can you describe the exact behaviour a bit more precisely? Did it hang without returning a result?

flyfish30:
This program has been successful compiled by ghc 8.0.1 and output is show in bellow.
   △    
  △ △   
 △   △  
△ △ △ △ 

In this playground, the file can not compiled.

rahulmutt:
I mean why wasn't it compiled? Did it show an error message from Eta or a Java runtime exception or did it just hang on compilation?

flyfish30:
The compile time is too long, the error message is show in bellow:
Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.ArrayIndexOutOfBoundsException: 4
at java.util.concurrent.FutureTask.report(FutureTask.java:122)
at java.util.concurrent.FutureTask.get(FutureTask.java:206)
at Sandbox.main(Sandbox.java:22)
Caused by: java.lang.ArrayIndexOutOfBoundsException: 4
at ghczmprim.ghc.CString$unpackzus18DO.enter(Unknown Source)
at ghczmprim.ghc.CString$satzus18FQ.thunkEnter(Unknown Source)
at eta.runtime.thunk.StgInd.enter(StgInd.java:19)
at eta.runtime.stg.StgClosure.evaluate(StgClosure.java:22)
at base.ghc.Base$zpzp.enter(Unknown Source)
at base.ghc.Base$satzusAD0.thunkEnter(Unknown Source)
at eta.runtime.thunk.StgInd.enter(StgInd.java:19)
at eta.runtime.stg.StgClosure.evaluate(StgClosure.java:22)
at base.ghc.io.handle.Text$zdwa8.enter(Unknown Source)
[...]

PR #350, opened 2017-05-05
Fix ArrayIndexOutOfBoundsException with unicode strings

pparkkin:
Fixes https://github.com/typelead/eta/issues/180.

### DIFF
--- a/libraries/ghc-prim/GHC/CString.hs
+++ b/libraries/ghc-prim/GHC/CString.hs
@@ -111,7 +111,7 @@ unpackCStringUtf8# str
   = unpack 0#
   where
     !bytes = getBytesUtf8# str
-    !len = strLength str
+    !len = alength# bytes
     unpack nh
       | isTrue# (nh ==# len) = []
       | isTrue# (ch `leChar#` '\x7F'#) = C# ch : unpack (nh +# 1#)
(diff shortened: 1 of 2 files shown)


pparkkin:
I was trying to figure out how to add some tests to `ghc-prim` for things like this, but couldn't figure out how to do that. Seems like without Prelude or much of anything, it's a little tough. Any ideas?

rahulmutt:
Great work! 

You can feel free to use Prelude as well. No need to test *just* ghc-prim since that's hard (you can't print anything in `ghc-prim`!). You can make an Etlas project that has `ghc-prim` and `base` in the `build-depends:` field and try out the following code:
```
{-# LANGUAGE MagicHash #-}

import GHC.CString
import GHC.Types

main :: IO ()
main = print $ unpackCStringUtf8# "SomeUnicodeCharHere"#
```
Just add a comment that it requires both `ghc-prim` and `base`, no need to submit the entire project!
The same bug might afflict some of the other functions that accept UTF8 strings - can you add tests for those as well?

pparkkin:
I couldn't figure out how to call `unpackCStringUtf8#` directly with unicode input. If I try to pass in e.g. `"△"#`, the build fails with a `primitive string literal must contain only characters <= '\xFF'`. I couldn't figure out another way to get a `JString#`.

So I just added a file to print out a bunch of UTF-8 characters.

rahulmutt:
Another way you can get around it:

```haskell
{-# LANGUAGE OverloadedStrings #-}
import GHC.Types
...
let (JS# str) = "SomeUnicodeHere" :: JString
print $ unpackUtf8String# str
```

PR #352, opened 2017-05-06
Change mkJString to toJString

Abhiroop:


rahulmutt:
Can you also remove the imports `GHC.Base` and `GHC.Pack`?

Issue #360, opened 2017-05-08
Update STM implementation to work with new concurrency model


rahulmutt:

rahulmutt:
Implemented in the `going-stackless` branch.

Issue #358, opened 2017-05-08
Make forkIO guarantee fairness for spawned tasks


rahulmutt:
If you run `forkIO` more than say N times, where N is the value of the `+RTS -N[N]` flag you send when you run the program, there is no guarantee that your forked threads' tasks will be given fairness in CPU time. If you use it less than N times, you are guaranteed that the tasks will run separately.

Let's fix this asap.

rahulmutt:
Implemented in the `going-stackless` branch. While we don't guarantee fairness exactly, we at least try to execute all idle TSOs in the global run queue after a small idle time.

PR #362, opened 2017-05-08
Wip/eta 338

### LINKED ISSUE #338: Make forkIO spawn a new Java thread
Eta should be able to handle at least basic concurrency without green threads as that has been deferred for after the first release. Using Eta in Apache Spark has shown that the Eta RTS can already work well in multi-threaded environments. The current logic for `forkIO` assumes the presence of green threads (modelled after GHC's implementation). `forkIO` should instead spawn a new Java Thread and attempt to obtain a free Capability to run the IO action that was passed in.

rahulmutt:
This primarily resolves #338, so Eta now has basic concurrency. 

Details:
- If you run `forkIO` more than say N times, where N is the value of the `+RTS -N[N]` flag you send when you run the program, there is no guarantee that your forked threads' tasks will be given fairness in CPU time. If you use it less than N times, you are guaranteed that the tasks will run separately. See #358.
- The threaded runtime is now default because no one writes useful non-concurrent applications anymore.
- MVar is implemented to work with the new `forkIO` change. STM primitives need to updated, see #360.
- A minor data structure change was made making the Eta stack use a `LinkedList` instead of a `Stack`. The result was a 20-30% increase in performance across the board, with an increase in GC allocation rate by up to 2x in some cases. This can be addressed with an intrusive list implementation, but for now the JVM GC is good enough to handle it.

[Before](https://circleci.com/gh/typelead/eta-benchmarks/45#tests/containers/0) and [After](https://circleci.com/gh/typelead/eta-benchmarks/46#build-timing/containers/0) (Click on `./eta-benchmarks.sh`)

Some sample entries here:

Benchmark | Before (ms) | After (ms)
--- | --- | ---
`imaginary/exp3_8` | 1018 | 718
`imaginary/wheel-sieve1` | 3585 | 2522
`spectral/lambda` | 9341 | 6542

### DIFF
--- a/libraries/base/GHC/Conc/Sync.hs
+++ b/libraries/base/GHC/Conc/Sync.hs
@@ -359,10 +359,9 @@ this value, use 'setNumCapabilities'.
 
 @since 4.4.0.0
 -}
-getNumCapabilities :: IO Int
-getNumCapabilities = do
-   n <- peek enabled_capabilities
-   return (fromIntegral n)
+foreign import java unsafe
+  "@static @field eta.runtime.stg.Capability.enabledCapabilities"
+  getNumCapabilities :: IO Int
 
 {- |
 Set the number of Haskell threads that can run truly simultaneously
@@ -383,7 +382,7 @@ setNumCapabilities i = c_setNumCapabilities (fromIntegral i)
 
 -- foreign import ccall safe "setNumCapabilities"
 c_setNumCapabilities :: CUInt -> IO ()
-c_setNumCapabilities = undefined
+c_setNumCapabilities = error $ "setNumCapabilities: unimplemented"
 
 -- | Returns the number of CPUs that the machine has
 --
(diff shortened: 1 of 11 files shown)


ashishnegi:
About the usage of `linked list vs vector` stacks, this link might be useful : http://agis.io/2014/03/25/contiguous-stacks-in-go.html 

The concept is called `stack thrashing`. 

Quoting page :
```
In such cases where the stack boundary happens to fall in a tight loop,
the overhead of creating and destroying segments repeatedly becomes significant. 
This is called the “hot split” problem inside the Go community. 
Rust folks call it “stack thrashing”.
```

So, benchmarking on this pathological case will be interesting.

rahulmutt:
Thanks for the share. Got me thinking about ways to optimise it even more! 

In the case of the Eta stack, alternate addition/removal doesn't seem to happen in practice as it does in Go, simply because most tail-recursive calls are optimised into loops hence stay within the same method and keep the stack relatively constant. In most cases you'll have a long string of additions followed by strings of removals and that doesn't happen nearly as often as it does in Go. The 30% increase was actually because of speeding up removes!

For the stack operations to be fast, we need O(1) insertions to the top *and* O(1) removals at arbitrary points (there are cases where just removing from the top is not enough). LinkedList clearly wins with such requirements, and Vector is slow since it does array copies on every removal.

PR #386, opened 2017-05-16
Add Fedora specific install instructions

jhrcek:
I encountered missing libs problems when building eta from sources on Fedora, so sharing my solution..


rahulmutt:
Thanks!

PR #394, opened 2017-05-21
Auto docker

AlexeyRaga:
## Changes

Build and publish Eta/Etlas docker container automatically.
This allows us always having a container with the latest version of Eta. This container can be conveniently used as a primary CCI build container for other Eta libraries (I'll add the instructions and an example later).

The container doesn't contain `~/.etlas/packages` so `etlas update` is expected (which would be expected anyway).

To enable auto publishing containers please set `DOCKER_EMAIL`, `DOCKER_USER` and `DOCKER_PASS` in CircleCI environment variables.

Container is pushed to docker hub as `typelead/eta:<git hash>`. 
When `master` is built, `typelead/eta:latest` is also published. Otherwies `typelead/eta:<branch-name>` is published.
If the commit is tagged, then `typelead/eta:tag` is also published.

See `docker/README.md` for more info.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,15 +17,20 @@ dependencies:
     - ~/.local
     - ~/.etlas
     - ~/.coursier
+    - ./.stack-work
+    - ./etlas/etlas/.stack-work
 
   pre:
     - mkdir -p ~/.local/bin
 
   override:
     - ./install.sh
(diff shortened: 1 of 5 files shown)


rahulmutt:
Looking good, thanks!

PR #400, opened 2017-05-25
latest jdk

AlexeyRaga:
## Changes

Oracle has removed the ability to download `JDK 8.92.14` and, as it seems, has changed the structure of the download URL. 
This PR is to use the latest `JDK 8.131.11` in a docker container.


rahulmutt:
Thanks! I think this will fix the Docker build.

PR #419, opened 2017-06-09
Make line buffering default for `stdout`

pparkkin:
I moved the `is_tty` stuff from `getCharBuffer` to where `mkHandle` was being called so I could pass in the buffering mode as a parameter to `mkHandle`.

After that, `mkUnBuffer` and `getCharBuffer` ended up being exactly the same, so I merged them into one.

And then of course pass in line buffering as default for `stdout`.

### DIFF
--- a/libraries/base/GHC/IO/Handle.hs
+++ b/libraries/base/GHC/IO/Handle.hs
@@ -639,7 +639,7 @@ dupHandle_ :: (IODevice dev, BufferedIO dev, Typeable dev) => dev
 dupHandle_ new_dev filepath other_side h_@Handle__{..} mb_finalizer = do
    -- XXX wrong!
   mb_codec <- if isJust haEncoder then fmap Just getLocaleEncoding else return Nothing
-  mkHandle new_dev filepath haType True{-buffered-} mb_codec
+  mkHandle new_dev filepath haType (BlockBuffering Nothing) mb_codec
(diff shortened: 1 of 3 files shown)


libraries/base/GHC/IO/Handle.hs
```diff
@@ -639,7 +639,11 @@ dupHandle_ :: (IODevice dev, BufferedIO dev, Typeable dev) => dev
 dupHandle_ new_dev filepath other_side h_@Handle__{..} mb_finalizer = do
    -- XXX wrong!
   mb_codec <- if isJust haEncoder then fmap Just getLocaleEncoding else return Nothing
-  mkHandle new_dev filepath haType True{-buffered-} mb_codec
+  is_tty <- IODevice.isTerminal new_dev
+  let buffer_mode
+         | is_tty = LineBuffering
+         | otherwise = BlockBuffering Nothing
+  mkHandle new_dev filepath haType buffer_mode mb_codec
```
rahulmutt: Another alternative to implement this is to make `Maybe BufferMode` as one of the argument to `mkHandle` and a value of `Nothing` should do the `isTerminal` check. I think doing it this way would make the code a lot more concise and avoid the repetitive `isTerminal` checks.

pparkkin: Great idea. Thanks!

Done.

PR #421, opened 2017-06-17
Update Dockerfile

aiya000:
I tried

```console
$ git clone https://github.com/typelead/eta
$ cd eta
$ docker build foo .
```

but I got the error of installing java :dog:

And, I think Dockerfile should get eta source from local.

I fixed it :dog:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Oracle stuff taken and modified from https://github.com/dockerfile/java/blob/master/oracle-java8/Dockerfile
 # MIT LICENSE of dockerfile/java  vvv
+
 #------------------------------------
 #The MIT License (MIT)
 #
@@ -39,6 +40,9 @@ ENV JAVA_VERSION_MAJOR=8 \
     PATH=$PATH:/root/.local/bin \
     LANG=C.UTF-8
 
+# Copy eta
+COPY ./ /usr/eta
+
 # Initialize apt
 RUN apt-get update -q
 
@@ -68,8 +72,10 @@ RUN apt-get install -q -y apt-file apt-utils && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /var/cache/oracle-jdk8-installer
 
-# Clone the latest eta
-RUN cd /usr && \
-    git clone --recursive https://github.com/typelead/eta && \
-    cd /usr/eta && \
+# Return to the latest eta
+RUN cd /usr/eta && \
+    git add -a && \
+    git stash && \
+    git checkout master && \
+    git pull && \
     ./install.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,8 @@ RUN apt-get install -q -y apt-file apt-utils && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /var/cache/oracle-jdk8-installer
 
-# Copy the current source of eta
-COPY ./ /usr/eta
-WORKDIR /usr/eta
-RUN ./install.sh
+# Clone the latest eta
+RUN cd /usr && \
+    git clone --recursive https://github.com/typelead/eta && \
+    cd /usr/eta && \
+    ./install.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,74 @@
-FROM ubuntu:xenial
-MAINTAINER Sibi Prabakaran <sibi@psibi.in>
+# Oracle stuff taken and modified from https://github.com/dockerfile/java/blob/master/oracle-java8/Dockerfile
+# MIT LICENSE of dockerfile/java  vvv
+#------------------------------------
+#The MIT License (MIT)
+#
+#Copyright (c) Dockerfile Project
+#
+#Permission is hereby granted, free of charge, to any person obtaining a copy
+#of this software and associated documentation files (the "Software"), to deal
+#in the Software without restriction, including without limitation the rights
+#to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+#copies of the Software, and to permit persons to whom the Software is
+#furnished to do so, subject to the following conditions:
+#
+#The above copyright notice and this permission notice shall be included in
+#all copies or substantial portions of the Software.
+#
+#THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+#AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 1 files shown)


rahulmutt:
Thanks! The java installation bit looks a *lot* cleaner. My main concern is removing the step where the eta repo is cloned. What if you wanted to pull the image straight from dockerhub, will it still work? Maybe you can add a step in the script that checks if the repo has already been cloned?

@AlexeyRaga Do any of your docker scripts assume that the Dockerfile will take care of downloading the eta repo?

aiya000:
Sorry, I misunderstood about copying eta repository !  
I fixed it :dog2:

aiya000:
We may can use below procedure instead my last commit 🤔 

```console
COPY ./ /usr/eta
RUN cd /usr/eta && \
	git add -a && \
	git stash && \
	git checkout master && \
	git pull && \
	./install.sh
```

Does @rahulmutt think about this ? :smile:

rahulmutt:
That works too I guess. Go ahead and make that change and I'll merge. If it doesn't work, I'll revert to your old commit.

aiya000:
Thank you :+1:

It is commited now :dog2:

AlexeyRaga:
@rahulmutt @aiya000 
In my opinion the proposed version has the following disadvantages:

- Not cleaning up Java installation will increase the container size. I really doubt that `/*src.zip`, `lib/missioncontrol`, `lib/visualvm`, `lib/*javafx*`, `jre/plugin`, `jre/bin/javaws`, `jre/bin/jjs`, etc. is benefitial for the compiler container.

- Doing `COPY ./ /usr/eta` and then building Eta within the container itself will cause the following effects:
  - The result container will have Eta source code (not needed for the compiler container)
  - After `./install.sh` happens, stuff like `.stack-works`, `~/.stack` will remain in the result container, blowing up its size _significantly_. For example, Eta's `.stack-work` eats `~1.5GB` of disk space, Etlas's `.stack-work` is ~200MB, `etlas-cabal` is about the same.

I can't justify making the compiler so huge from the top of my head as I generally prefer containers to have only stuff that is needed for what they intend to do, and to keep them as small as possible.

rahulmutt:
@AlexeyRaga It would be great to get a PR optimising the size of the image.

From what you said, we just need to add extra steps removing the `~/.stack` and `./.stack-work` folders correct?

AlexeyRaga:
Theoretically yes, we should remove `~/.stack` after the installation, and, perhaps, the whole source code folder (because why having it there?)
The problem is that you can't simply do the former after `COPY` instruction: files will be removed from the current layer, but will still remain in the previous because of how Docker works.
We could avoid it by pulling it from Git and building in one instruction, we just need to figure out the hash that is being built.

Also building it in CI within the container would mean that we build it twice (implies time).

Depends on what problem we are trying to solve here. Building a container locally for some reason? This could be done by simply adding another (dev) Dockerfile. 
Another way of doing it would be to:

- Build Eta locally in a way CI does:
```
    $ stack --docker --docker-image=fpco/stack-build:latest setup
    $ stack --docker --docker-image=fpco/stack-build:latest --no-docker-set-user exec bash
    stack# ./install.sh
    
    stack# mkdir -p ./docker/bin/.local 
    stack# cp -r ~/.local/bin ./docker/bin/.local/
    stack# cp -r ~/.eta ./docker/bin/
    stack# cp -r ~/.etlas ./docker/bin/
    stack# rm -rf ./docker/bin/.etlas/logs/*
    stack# rm -rf ./docker/bin/.etlas/packages/*

    stack# exit

    $ ./docker/circleci-build-container.sh
```

It also can be scripted into something like `./docker/circleci-build-container.sh dev-build`...

Anyway, do whatever you think is convenient :)

rahulmutt:
@aiya000 We still have a need for a dev Dockerfile - can you rename this to Dockerfile.dev and make sure your base image is `openjdk:8-jdk-slim`? If your build is failing, take a look at [this](https://github.com/typelead/eta/blob/master/Dockerfile) Dockerfile for hints on how to fix it.

aiya000:
@rahulmutt @AlexeyRaga  Sorry for waiting x(

@AlexeyRaga I think it is OK to removing 😄 

- /*src.zip
- lib/missioncontrol
- lib/visualvm
- lib/*javafx*
- jre/plugin
- jre/bin/javaws
- jre/bin/jjs

I can add doing it to DockerFile, but I didn't know it is not really needed :bow: .  
Should I add it to DockerFile ?

AlexeyRaga:
@aiya000 if you use `openjdk:8-jdk-slim` (as `docker/Dokerfile` does) then you won't need to install Java yourself.

And if after the installation is completed you remove stuff that is not needed (like sources folder and, maybe, `.stack`) within the same layer then it would be awesome.

aiya000:
> Doing COPY ./ /usr/eta and then building Eta within the container itself will cause the following effects:
> 
> - The result container will have Eta source code (not needed for the compiler container)
> - After ./install.sh happens, stuff like .stack-works, ~/.stack, etc. will remain in the result container, blowing up its size significantly. For example, Eta's .stack-work eats ~1.5GB of disk space, Etlas's .stack-work is ~200MB, etlas-cabal is about the same.

It is bad definitely x(  
I'll optimize DockerFile for it :+1:

aiya000:
> @aiya000 We still have a need for a dev Dockerfile - can you rename this to Dockerfile.dev and make sure your base image is openjdk:8-jdk-slim? If your build is failing, take a look at this Dockerfile for hints on how to fix it.

OK :+1:

Are these best to pursue [this](https://github.com/typelead/eta/blob/master/Dockerfile) and commit as Dockerfile, or rename Dockerfile.dev ?

aiya000:
mmm...  
I think [this](https://github.com/typelead/eta/blob/master/Dockerfile) is completed.  
May this PR is not needed 🤔  ?

rahulmutt:
The problem with that Dockerfile is that is makes no use of caching so when you are building small changes, it rebuilds everything from scratch! Moreover, it downloads from github instead of using the local directory as a build context which doesn't make sense if you're building a branch other than master. Your Dockerfile makes very good use of caching, so with a couple changes, it can make for a perfect dev Dockerfile.

AlexeyRaga:
When using local caching please consider that local stuff can be built for a different OS (MacOS for example) and cannot be directly put into the docker container.

This is one of the reasons I mentioned `stack --docker --docker-image=fpco/stack-build:latest` when building, this way on any OS `stack` will build linux binaries and will cache them locally under `.stack-work/x86_64-linux-*` so _these_ binaries can be reused instead of building from scratch.

aiya000:
@rahulmutt @AlexeyRaga I understood the problem 😃  
Can I `git-rebase` this PR to the current HEAD of master with `git-push -f` ?

rahulmutt:
Just so we're clear on the use case for the dev Dockerfile:
- This is for people who are hacking on the Eta compiler itself or want to keep up with the bleeding edge without downloading entirely new images. Not sure how the setup would look like, but something that uses caching in the optimal way such that using an update build context triggers the smallest possible rebuild (if that's even possible).

@AlexeyRaga Thanks for the correction.
@aiya000 It's probably better to close this pull request and start a new one from the tip of master.

rahulmutt:
@aiya000 Any updates on this?

aiya000:
@rahulmutt Very very sorry !!
I'll open new PR for HEAD,  
and I'll update this in new PR,  
in today and tomorrow 😭

aiya000:
@rahulmutt

I thought once more.  
But I think the caching cannot be not enabled,
because a point for the caching is already removed,  
I think that point is needed by the caching.

- [eta/Dockerfile at master · typelead/eta · GitHub](https://github.com/typelead/eta/blob/master/Dockerfile#L19)

Sorry... , how do you think ?

PR #432, opened 2017-07-29
CCI 2.0

AlexeyRaga:
# Migrate to `CircleCI 2.0`

### Motivation
CircleCI 2.0 has been released and 1.0 is deprecated.
The benefits of using CircleCI 2.0 compare to 1.0 include: much faster builds, much more convenient and flexible caching, custom build images.

### Changes
- Migrate to CCI 2.0
- Using official `stack` build image `fpco/stack-build`
- Better caching strategy (fallback caches)
- Introduced build workflow (see below)

### Workflow
CircleCI 2.0 supports [Workflows](https://circleci.com/docs/2.0/workflows/) which allow splitting the build process into semi-independent parts.

The workflow proposed in this PR (called `full-cycle`) does:
- Run the main `eta` build process
- When `build` succeeds, run the following steps _in parallel_:
  - Runs `eta` tests (which includes building tons of packages and currently has a long history of failures)
  - Build/release Eta/Etlas docker container (release will only happen for the `master` branch)
  - Triggers build on other projects (only happens in `master`)

The workflow is defined as:

```
  full_cycle:
    jobs:
      - build

      - test:
          requires:
            - build

      - release_container:
          requires:
            - build
          filters:
            branches:
              only: master

      - dev_container:
          requires:
            - build
          filters:
            branches:
              ignore: master

      - trigger_builds:
          requires:
            - build
          filters:
            branches:
              only: master
```

### Room for improvements
- The build process can be improved further by figuring out better integration between caches and git submodules. I have never used git submodules in my projects (I generally prefer publishing and reusing artifacts instead) so it is not done for this build.


rahulmutt:
Looking good! I'll merge as soon as the build fails at a similar point as the previous failures.

AlexeyRaga:
These `test` steps are failing for a long time before my PR (checked CCI history). Looks like `eta` cannot build one of the sources of one of the packages.

Please correct me if I am wrong and I broke something, because I could ;)

rahulmutt:
No mistake on your side. We need to update some of the patches on `eta-hackage`, which should fix it. Thanks for the upgrade! Merging.

PR #440, opened 2017-08-05
Make rts-clean works in windows

jneira:
Installing rts doesnt seem to change the hash of the lib in the etlas libs dir


rahulmutt:
Great thanks! :)

PR #460, opened 2017-08-18
Build Openjdk Slim based container

AlexeyRaga:
## Changes
- Use `openjdk-slim` container instead of Ubuntu+Java8. This reduces the result container size from `521MB` to `177MB`.
- Build docker container in branches. `typelead/eta:latest` tag will be pushed from `master`, `typelead/eta:<branchName>` will be pushed from dev branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
             - ~/.stack
 
       - restore_cache:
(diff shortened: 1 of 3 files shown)


rahulmutt:
Awesome stuff! Like that you are building images for `dev` branches as well.

PR #462, opened 2017-08-19
fix #454

### LINKED ISSUE #454: eta: panic! (the 'impossible' happened)         InvalidAbsFile "..
when there is a  -i option that has references the parent directory ".." : 

```
tests# eta  -i../src Test.hs
[1 of 7] Compiling Transient.Internals ( ../src/Transient/Internals.hs, ../src/Transient/Internals.jar ) [flags changed]
eta: panic! (the 'impossible' happened)
  (Eta version 0.0.9b2):
        InvalidAbsFile "../src/Transient/Internals.jar"
```

agocorona:
Checked with my use cases.

### DIFF
--- a/compiler/ETA/Utils/JAR.hs
+++ b/compiler/ETA/Utils/JAR.hs
@@ -50,7 +50,7 @@ import Path
 import Path.IO (copyFile)
 import Data.Map.Lazy (keys)
 import Data.Map.Strict (filterWithKey)
-import Data.List (sortBy)
+import Data.List (sortBy,isPrefixOf )
 import System.Directory hiding (copyFile)
 import Data.Text (Text)
 import qualified Data.Text as T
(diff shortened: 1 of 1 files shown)


rahulmutt:
Looks good. Let's see how the build goes.

Also, we may hit [this](http://neilmitchell.blogspot.co.uk/2015/10/filepaths-are-subtle-symlinks-are-hard.html) problem in the future, but I think it won't be too much of a problem for the use case of source code for projects.

agocorona:
For that reason probably  `canonicalizePath` actually acess to the path, to determine the real parent directory (I guess)

rahulmutt:
Hmm yes, but canonicalizePath doesn't work for non-existent paths. Hmm I wonder if there's a way to normalise non-existent paths by stripping of the right side of a path until you get an existant path and then slap on the part your stripped off.

What do you think? Should we be worried about that obscure case?

agocorona:
mmm  now I remember that canonicalizePath is not capable of elimianting ".."  even if the file exist. At least in my case.  I think that this only hit in rare cases and even in these cases the programmer has the option to write the path without .. if it points to a symbolic link

rahulmutt:
Ok cool. I'll merge then.

Issue #189, opened 2017-01-19
Generate line number information in class files


rahulmutt:
In stack traces, any code generated by Eta has `(Unkown Source)` listed as the line number since Eta doesn't generate any info right now. The information is available at the STG level via `StgTick`. This information should be used to generate the appropriate class file attributes that can be used to make debugging a whole lot easier.

rahulmutt:
I've outlined the steps to get this implemented below. This is what I feel is the most immediate approach to getting this feature in. There may be alternatives that you can choose to pursue based on your style of programming.

1. Add functionality in `codec-jvm` for generating the [LineNumberAttribute](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.12) attribute.
  a. In `Codec.JVM.Attr`, a new constructor for the `Attr` type should be added, probably something along the lines of `ALineNumber [(Offset, LineNumber)]` where `LineNumber` can be a `newtype` or a `type` synonym.
  b. Update `attrName` to handle the new case.
  c. Update `putAttrBody` to handle the new case.

2. Add functionality in `codec-jvm` for generating the [SourceFileAttribute](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.10) attribute.
  a. In `Codec.JVM.Attr`, a new constructor for the `Attr` type should be added, probably something along the lines of `ASourceFile Text`.
  b. Update `attrName` to handle the new case.
  c. Update `putAttrBody` to handle the new case.
  d. Update `unpackAttr` to handle the new case. 
        i. This is required since `SourceFile` references a constant in the constant pool for the name. 

3. In `codec-jvm`, provide APIs to add attributes for methods.
  a. Update `Codec.JVM.ASM.MethodDef` to store attributes.
  b. Update `Codec.JVM.ASM.mkMethodDef` to handle the change in (a).
  c. Add a new function `addAttrsToMethodDef` which will allow you to update an existing `MethodDef` and add new attributes to it.
  d. Update `Codec.JVM.Method.MethodInfo` to store attributes.
  e. Update `Codec.JVM.ASM.mkClassFile` to convert between `MethodDef` and `MethodInfo` properly handling the new attributes-storing fields in each type.

4. In `Codec.JVM.ASM.Code.Instr`, add the facility to record line number info.
  a. Augment `InstrState` with a field for storing line number information.
  b. Add a new function `getBCSL` that is similar to `getBCS` which returns the line number information from (a).
[...]

PR #492, opened 2017-09-01
Emitting line numbers in code generation

jneira:
[#189] Some considerations:
* it adds some overhead to cgExpr when no SourceNote's are generated (without -g flag?): it checks the list of line numbers for every cgExpr evaluation
* the name of the source file is put in every SourceNote expr, when should be the same for the class being generated
* the name of source file contains the relative os-specific path (f.e. `.\GHC\Err.hs`). We have to determine what should be the best name. If we follow the [jvm bytecode spec](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.10):
> The string referenced by the sourcefile_index item will be interpreted as indicating the name of the source file from which this class file was compiled. It will not be interpreted as indicating the name of a directory containing the file or an absolute path name for the file; such platform-specific additional information must be supplied by the run-time interpreter or development tool at the time the file name is actually used. 

... the name should be Err.hs but it is less helpful than indicating the full module name in some way

I've compiled base with the version and i've got this stracktrace:
```
D:\ws\eta\eta\examples>java -jar RunHelloWorld.jar
Hello World!
Exception in thread "main" eta.runtime.exception.EtaException: Testing stacktraces
        at base.ghc.Err$error.enter(.\GHC\Err.hs:35)
        at main.Main$$Ls1LWsat.thunkEnter(HelloWorld.hs:5)
        at eta.runtime.thunk.CAF.enter(CAF.java:31)
        at eta.runtime.thunk.Thunk.applyV(Thunk.java:121)
        at base.ghc.Base$thenIO1.enter(./GHC/Base.hs:1073)
        at eta.runtime.apply.PAP.apply(PAP.java:31)
        at eta.runtime.apply.PAP.applyV(PAP.java:41)
        at eta.runtime.stg.Closures$EvalLazyIO.enter(Closures.java:100)
        at eta.runtime.stg.Capability.schedule(Capability.java:153)
        at eta.runtime.stg.Capability.scheduleClosure(Capability.java:100)
        at eta.runtime.Runtime.evalLazyIO(Runtime.java:189)
        at eta.runtime.Runtime.main(Runtime.java:182)
        at eta.main.main(Unknown Source)
```
* I've run the verify script (the new windows version) and it seems to check the class

### DIFF
--- a/compiler/ETA/CodeGen/Expr.hs
+++ b/compiler/ETA/CodeGen/Expr.hs
@@ -5,6 +5,7 @@ import ETA.Types.Type
 import ETA.Types.TyCon
 import ETA.BasicTypes.Id
 import ETA.BasicTypes.VarEnv
+import ETA.BasicTypes.SrcLoc
 import ETA.Prelude.PrimOp
 import ETA.StgSyn.StgSyn
 import ETA.BasicTypes.DataCon
@@ -27,35 +28,49 @@ import Codec.JVM
 
 import Data.Monoid((<>))
 import Data.Foldable(fold)
-import Data.Maybe(mapMaybe)
+import Data.Maybe(mapMaybe,maybe)
 import Control.Monad(when, forM_, unless)
 
 cgExpr :: StgExpr -> CodeGen ()
-cgExpr (StgApp fun args) = traceCg (str "StgApp" <+> ppr fun <+> ppr args) >>
+cgExpr expr = do
+  expr' <- preCgExpr expr
+  doCgExpr expr' 
+
+preCgExpr :: StgExpr -> CodeGen StgExpr
+preCgExpr e@(StgTick _ _) = return e 
+preCgExpr expr = do
+  mbLn <- getInnermostLineNumber
+  let ln = fmap mkLineNumber mbLn
+  traceCg (str $ "Emitting line number: " ++ show ln)
+  emit $ maybe mempty emitLineNumber ln
+  resetLineNumbers
+  return expr
+  
+doCgExpr :: StgExpr -> CodeGen ()
+doCgExpr (StgApp fun args) = traceCg (str "StgApp" <+> ppr fun <+> ppr args) >>
                            cgIdApp fun args
-cgExpr (StgOpApp (StgPrimOp SeqOp) [StgVarArg a, _] _) = cgIdApp a []
-cgExpr (StgOpApp op args ty) = traceCg (str "StgOpApp" <+> ppr op <+> ppr args <+> ppr ty) >>
+doCgExpr (StgOpApp (StgPrimOp SeqOp) [StgVarArg a, _] _) = cgIdApp a []
+doCgExpr (StgOpApp op args ty) = traceCg (str "StgOpApp" <+> ppr op <+> ppr args <+> ppr ty) >>
                                cgOpApp op args ty
-cgExpr (StgConApp con args) = traceCg (str "StgConApp" <+> ppr con <+> ppr args) >>
+doCgExpr (StgConApp con args) = traceCg (str "StgConApp" <+> ppr con <+> ppr args) >>
                               cgConApp con args
 
-cgExpr (StgTick t e) = do
-  traceCg (str "StgTick" <+> ppr t)
+doCgExpr (StgTick t e) = do
+  traceCg (str "StgTick, tickish:" <+> ppr t)
   cgTick t
   cgExpr e
-
   
-cgExpr (StgLit lit) = emitReturn [mkLocDirect False $ cgLit lit]
-cgExpr (StgLet binds expr) = do
+doCgExpr (StgLit lit) = emitReturn [mkLocDirect False $ cgLit lit]
+doCgExpr (StgLet binds expr) = do
   forbidScoping (cgBind binds)
   cgExpr expr
-cgExpr (StgLetNoEscape _ _ binds expr) =
+doCgExpr (StgLetNoEscape _ _ binds expr) =
   cgLneBinds binds expr
 
-cgExpr (StgCase expr _ _ binder _ altType alts) =
+doCgExpr (StgCase expr _ _ binder _ altType alts) =
   traceCg (str "StgCase" <+> ppr expr <+> ppr binder <+> ppr altType) >>
   cgCase expr binder altType alts
-cgExpr _ = unimplemented "cgExpr"
+doCgExpr _ = unimplemented "cgExpr"
 
 cgLneBinds :: StgBinding -> StgExpr -> CodeGen ()
 cgLneBinds (StgNonRec binder rhs) expr = do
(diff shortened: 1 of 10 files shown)


puffnfresh:
Awesome! :fireworks: 

The spec seems a bit disappointing but I think we should follow it.

jneira:
@puffnfresh agree, the context to avoid clashes could be extracted from class package name anyway

rahulmutt:
Awesome work @jneira! Probably one of the most important patches till date :) 

I agree with @puffnfresh that we should just go with the spec (so `Err.hs`) above. Moreover, we've done the encoding of the class name to store the entire module name (lowercased) anyways, so it's not such a big deal what we put for SourceFile. The line number was the most important aspect.

 I do have a better solution beyond this:
Customise the exception printing to make use of that extra information via overriding `printStackTrace()` for `EtaException` to give much more accurate information (like column number ranges). I'll file an issue for that, but for now this should be an order of magnitude more helpful than `Unknown Source` :)

rahulmutt:
By the way, can you make `-g` flag default and let me know if compilation speed is significantly slower? We need this as a default if it's to be useful.

compiler/ETA/CodeGen/Expr.hs
```diff
@@ -319,3 +341,11 @@ cgAlgAltRhss binder alts = do
       allBranches = [ (getDataConTag con, code)
                     | (DataAlt con, code) <- taggedBranches ]
   return (maybeDefault, branches)
+
+cgTick :: Tickish Id -> CodeGen ()
+cgTick (SourceNote srcSpan _) = do
+  let srcLoc = realSrcSpanStart srcSpan
+  traceCg (str $ "StgTick , tickish: SourceNote, srcSpan: " ++ show srcSpan)
+  addLineNumber $ srcLocLine srcLoc
+  setSourceFileName $ fastStringText $ srcLocFile srcLoc
```
rahulmutt: `setSourceFileName` should ideally do some sort of filtering. Because according to this logic, the last SourceNote in a given binding will determine the SourceFile and I don't think this will always be the current module (due to inlining).

rahulmutt:
Thanks a lot for converting the verify script to a cmd script - will be of great help to the Windows devs.

rahulmutt:
Ok that fixes it, thanks for the quick response! So all that's remaining is making `-g` default - hopefully it won't have a *visible* impact on compilation time.

jneira:
Yeah, i think this version is better although it has needed to change `HSCMain`

I'll do some tests to comapre compilation times

rahulmutt:
Test something like `containers` which already takes a while - the difference should be apparent if there is any.

jneira:
My tests dont show significant differences between versions (10 samples for version):
```
eta line-number version

without -g
Avg: 56675.5187ms
Min: 56058.9607ms
Max: 57895.6654ms

with -g

Avg: 61082.3639ms
Min: 60428.5235ms
Max: 61912.3757ms

eta master version

without -g 10 samples
Avg: 56290.0983ms
Min: 55729.9257ms
Max: 57162.3847ms

with -g
Avg: 61905.0487ms
Min: 60268.0346ms
Max: 64600.9394ms
```
* I think we should set this as default in some way but maybe at etlas level, letting the user set off the -g flag

rahulmutt:
Agreed, go ahead and set it as a default for now. Debugging info is pretty important, so we'll figure out ways to decrease the compile times in other ways later on.

compiler/ETA/CodeGen/Expr.hs
```diff
@@ -319,3 +341,9 @@ cgAlgAltRhss binder alts = do
       allBranches = [ (getDataConTag con, code)
                     | (DataAlt con, code) <- taggedBranches ]
   return (maybeDefault, branches)
+
+cgTick :: Tickish Id -> CodeGen ()
+cgTick (SourceNote srcSpan _) = do
+  let srcLoc = realSrcSpanStart srcSpan
+  addLineNumber $ srcLocLine srcLoc
```
rahulmutt: I just realised there's a tricky corner case here. You should only addLineNumber if the file corresponds to the current module, otherwise you'll get a line number that doesn't match the source file.

jneira: Ok, just out of curiosity, do you know the cause of that case?

jneira:
After some comments with @rahulmutt, we thought that the best option would be to set eta `-g` option active by default. In this way eta generated bytecode will have by default the same attrs than `javac` generated one: src file and line numbers. To let users unset it, the chosen option has been mimic the actual [ghc behaviour](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/flags.html#code-generation-options) (`-g⟨n⟩`) so
* `-g=0` unsets the debug flag to get bytecode without source file and line numbers attrs
* The rest of cases (`-g`,`-g=1`,etc) sets the debug flag active

compiler/ETA/CodeGen/Expr.hs
rahulmutt: Here's an example:

`GHC.List`
```
creating new closure...errorEmptyList
StgTick errorEmptyList <GHC/List.hs:(1010,1)-(1011,48)>
StgTick build <GHC/Base.hs:785:1-18>
StgLet sat_sNKV
```

In this example, the line number would show up as 785 with source file as `GHC/List.hs` when it should be 1010.

jneira: Aja, i see, it generates ticks for all the call stack, including external modules ones

jneira:
@rahulmutt do you mean [this one](https://github.com/typelead/eta/pull/492/files/c5bc156e5e1bf1333d0f4f1fc3a1e7aacc118df8#r136722195) or another one? i think [this commit](https://github.com/typelead/eta/pull/492/commits/3ef5b54f16308f33e4a8a3fab4bb6b56eacccbfd#diff-4ba0e63720b82a17f91378e398c31fe2) should handle it

compiler/ETA/CodeGen/Expr.hs
```diff
@@ -342,8 +342,13 @@ cgAlgAltRhss binder alts = do
...
+cgTick :: Tickish a -> CodeGen ()
 cgTick (SourceNote srcSpan _) = do
-  let srcLoc = realSrcSpanStart srcSpan
-  addLineNumber $ srcLocLine srcLoc
+  mbCgSrcFile <- getSourceFileName
+  case mbCgSrcFile of
+    Just cgSrcFile -> do
+      let srcLoc  = realSrcSpanStart srcSpan
+          srcFile = fastStringText $ srcLocFile srcLoc
+      when (srcFile == cgSrcFile) $ addLineNumber $ srcLocLine srcLoc
```
rahulmutt: Will this condition hold for hierarchical modules?

From how you've defined it, for `GHC/List.hs`, `mgCgSrcFile` will be `Just "List.hs"` and `srcLocFile` will show up as `GHC/List.hs`. I think you'll need to store the *full* source file name as well in the CgState and do comparison against that.

jneira: Ouch, you are right, i changed the way to get the source file just for this reason.
I'll fix asap

rahulmutt:
Looks great now. Great work @jneira! Will merge after CI passes.

rahulmutt:
@jneira Just for reference, this change made the test build run longer by just 4 minutes (1 hour 9 minutes vs 1 hour 13 minutes). Not sure how that matches up with the numbers you shared, but it looks like the impact is not something we need to worry about.

PR #501, opened 2017-09-06
Simplify handling line numbers of StgTicks

jneira:
After talking with @rahulmutt about the posibility of simplyfing the handling of `StgTick` in #492 , i've implemented another solution replacing mutable state  in `CodeGen` and the splitting of `cgExpr` between `preCgExp` and `doCgExpr` with an inner loop in `stgTick`.

### DIFF
--- a/compiler/ETA/CodeGen/Expr.hs
+++ b/compiler/ETA/CodeGen/Expr.hs
@@ -325,28 +325,22 @@ cgAlgAltRhss binder alts = do
   return (maybeDefault, branches)
 
 cgTick :: StgExpr -> CodeGen ()
-cgTick = cgTickWithLn 0 
-  where cgTickWithLn n (StgTick tickish subExpr) = do
-          traceCg (str "StgTick, tickish:" <+> ppr tickish)
-          nxtLn <- nextLine tickish n
-          cgTickWithLn nxtLn subExpr
+cgTick expr = do
+  mbCgFilePath <- getSourceFilePath
+  let srcLoc = realSrcSpanStart
+      isModuleTick (SourceNote srcSpan _) =
+          let srcFile = unpackFS $ srcLocFile $ srcLoc srcSpan
+          in  maybe False (equalFilePath srcFile) mbCgFilePath
+      isModuleTick _ = False
+      (ticks,subExpr) = stripStgTicksTop (not . tickishIsCode) expr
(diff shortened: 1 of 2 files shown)


jneira:
i've detected a strange behavior in ticks:
* With this code:
```haskell
main = do
  putStrLn "Hello World!"
  error "Testing stacktrace"
```
* The call to `error` doent have StgTick associated (and no line number)
  * So the line shown in stacktrace is the last emitted (the start of main method)
* With
```haskell
main = do
  putStrLn "Hello World!"
  error "Testing stacktrace"
  return ()
```
* the `error` and `return` calls have their own StgTicks (and line numbers) and the stacktrace shows correctly the line

jneira:
As @rahulmutt suggested we could use ` stripStgTicksTop` to get the consecutive ticks instead of raw recursion. The vesion could be:
```haskell
cgTick :: StgExpr -> CodeGen ()
cgTick expr = do
  mbCgFilePath <- getSourceFilePath
  let srcLoc = realSrcSpanStart
      isModuleTick (SourceNote srcSpan _) =
          let srcFile = unpackFS $ srcLocFile $ srcLoc srcSpan
          in  maybe False (equalFilePath srcFile) mbCgFilePath
      isModuleTick _ = False
  case  stripStgTicksTop isModuleTick expr of
    ((SourceNote srcSpan _:_,_)) -> do
      let n = srcLocLine $ srcLoc srcSpan
      traceCg (str $ "Emitting line number: " ++ show n)
      emit $ emitLineNumber $ mkLineNumber n        
    _ -> return ()
```

do you like it more than actual one? it has a diff: the sucesive ticks are not traced, only the first one

rahulmutt:
Regarding the issue with `error` not generating ticks, I think we should file a separate issue to track it. I'm not sure how prevalent that will be in real applications, but we'll see.

I do like the `stripStgTicksTop` because it's more readable. And yes, please loop over them and print out the ticks with `traceCg` so that we can clearly see all the ticks that show up - it makes it easier to debug. Btw are you still following the innermost selection rule? Because it looks like the code above selects the outermost.

rahulmutt:
And I'm wondering whether the recursive call to `cgExpr` should be placed inside `cgTick` and the argument should be the second component of the tuple returned by `stripStgTicksTop`.

jneira:
@rahulmutt i've pushed a fixed version using `stripStgTicksTo`. I dont test its performance but it should be faster than the previous version

Issue #486, opened 2017-08-29
NoClassDefFound error for generated class


JustusAdam:
In our project we define a custom monad transformer

```hs
newtype OhuaT m a = OhuaT { runOhuaT' :: RWST CompilerEnv Errors CompilerState m a }
    deriving (Functor, Applicative, Monad, MonadIO, MonadTrans)
```

And an associated class

```hs
class Monad m => MonadOhua m where
    recordError :: Error -> m ()
    modifyState :: (CompilerState -> CompilerState) -> m ()
    getState :: m CompilerState

instance Monad m => MonadOhua (OhuaT m) where
    recordError err = OhuaT $ tell [err]
    modifyState = OhuaT . modify
    getState = OhuaT get
```

Plus a function to initialize the state and run the thing 

```hs
runOhuaT0 :: Monad ctxt => OhuaT ctxt result -> HS.HashSet Binding -> ctxt result
runOhuaT0 f taken = do
    (val, errors) <- evalRWST (runOhuaT' f) (error "Ohua has no environment!") (CompilerState nameGen 0)
    unless (null errors) $ error $ intercalate "\n" errors
    return val
  where
    nameGen = initNameGen taken
```

This all compiles but if i try to run it with even a simple computation

```hs
main = print =<< runOhuaT0 generateBinding mempty
```

I get the following error:

```
Exception in thread "main" java.lang.NoClassDefFoundError: ohua_core/ohua/Monad$$fMonadOhuaT
        at main.Main$main1.enter(Unknown Source)
        at main.Main$DZCmain.enter(Unknown Source)
        at eta.runtime.apply.Function.applyV(Function.java:16)
        at eta.runtime.stg.Closures$EvalLazyIO.enter(Closures.java:96)
        at eta.runtime.stg.Capability.schedule(Capability.java:150)
        at eta.runtime.stg.Capability.scheduleClosure(Capability.java:97)
        at eta.runtime.Runtime.evalLazyIO(Runtime.java:189)
        at eta.runtime.Runtime.main(Runtime.java:182)
        at eta.main.main(Unknown Source)
Caused by: java.lang.ClassNotFoundException: ohua_core.ohua.Monad$$fMonadOhuaT
        at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
[...]

PR #502, opened 2017-09-06
Adding undersores for class dict instances

JustusAdam:
First of hopefully two pull requests to address the issue in #486. 

This changes the generation function for class instance dicts.

Class instances

```hs
class MyClass a b c where

instance MyClass Tt Aa b where
instance MyClass Tt A ab where
```

Would previously both result in the instance dict named `$fMyClassTtAab`.
To reduce name clashes and make instances more readable we introduce a `_` as separator for individual type arguments resulting in the two dicts `$fMyClass_Tt_Aa_b` and `$fMyClass_Tt_A_ab`.

## Dicussion

Considering that we need further changes to address the case-sensitivity issue do we want to escape literal `_`?

Aka should 

`MyClass T_t` be `$fMyClass_T_t` (no escaping)
or 
`MyClass T_t` be `$fMyClass_T__t` (`_` escaped to `__`)?

### DIFF
--- a/compiler/ETA/TypeCheck/TcEnv.hs
+++ b/compiler/ETA/TypeCheck/TcEnv.hs
@@ -1,6 +1,6 @@
 -- (c) The University of Glasgow 2006
 
-{-# LANGUAGE CPP, FlexibleInstances #-}
+{-# LANGUAGE CPP, FlexibleInstances, LambdaCase #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}  -- instance MonadThings is necessarily an orphan
 
 module ETA.TypeCheck.TcEnv(
@@ -94,6 +94,7 @@ import ETA.Utils.Util
 import ETA.Utils.Maybes( MaybeErr(..) )
 import Data.IORef
 import Data.List
+import Data.Foldable
 
 #include "HsVersions.h"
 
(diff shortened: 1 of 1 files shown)


rahulmutt:
The build failed with this:
```
java.lang.NoSuchMethodException: base.java.Exception.$fExceptionJException()
	at java.lang.Class.getMethod(Class.java:1786)
	at eta.runtime.stg.Closures.loadClosure(Closures.java:63)
	at eta.runtime.stg.Closures.<clinit>(Closures.java:50)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at Verify.main(Verify.java:39)
```
Can you also change `eta.runtime.stg.Closures` to reference `$fException_JException` instead of `$fExceptionJException`? Also run `tests/verify/verify.sh` to ensure that this change works.

rahulmutt:
And please do escaping because `_` is an artifact of the encoding we're doing and `_` itself should be able to be distinguished from the separator `_`.

compiler/ETA/TypeCheck/TcEnv.hs
```diff
@@ -784,12 +785,26 @@ newDFunName :: Class -> [Type] -> SrcSpan -> TcM Name
...
-        ; let info_string = occNameString (getOccName clas) ++
-                            concatMap (occNameString.getDFunTyKey) tys
-        ; dfun_occ <- chooseUniqueOccTc (mkDFunOcc info_string is_boot)
+        ; dfun_occ <- chooseUniqueOccTc $ mkDFunOcc (mkInfoString clas tys) is_boot
         ; newGlobalBinder mod dfun_occ loc }
 
 {-
+It may be advantageous at some point to build a string for a type which actually
+encompasses the structure of the type like parethesizing.
+Even in the new mkInfoString something like `a (b c)` and `(a b) c` will result
```
rahulmutt: You can handle parenthesis with Z-encoding, but I'm not sure it's worth it at this point. We can do that if it seriously becomes an issue.

JustusAdam:
Okay, interesting. For me this morning for some reason the build worked and also the verify script. It didn't a few days back when I tried it. I suppose something is wrong with my test setup, I'll have to figure that out.

I also considered some kind of encoding for parentheses, but I couldn't think of one that I would have deemed readable.

rahulmutt:
You can run `./cleaninstall.sh` to ensure that the old version is cleanly removed. Another issue may be if you did the binary installation first, in which case `etlas` would be using your downloaded binary and not the source installed `eta`.

Yup, as I said, nothing to worry about (in regards to the parenthesis).

rahulmutt:
Hey @JustusAdam, let me know if you have time to fix the naming change in the runtime as well. If not, I'll merge this and immediately push the fix.

JustusAdam:
Oh if you want you can fix it. In fact you can fix it on this branch. I have enabled `allow edits from maintainers` so you should be able to push directly into this PR.
I might get around to fixing it today but I am not sure.

rahulmutt:
Merged, thanks for bringing this in!

PR #614, opened 2018-01-18
Added default behavior during ceration of new file channels on Windows platform

indegro:


indegro:
@rahulmutt I added checks during creation of new file channel. 
Default behavior is to check if file system supports posix permission, if not then create without it.
There may be a case when file store may support both posix and acl permission, but it a little bit complex case. jdk has an option to do this. 

I propose to create a separate ticket for that case.

ps.
Case(when file store supports both acl and posix) is possible on unix platforms. 
But it is a little bit rate case at the moment as far as I know.

rahulmutt:
This looks good so far and will work great for Unix systems. But this may yield incorrect behavior on Windows (i.e. not setting the appropriate file permissions when they should).

I think for Windows you can fall back on the pre-Java 7 `java.io.File.set(Executable|Readable|Writeable)` APIs.

indegro:
@rahulmutt As I see It is possible to translate posix owner/other permissions but not group. Is it ok?

rahulmutt:
Yes that's fine. You can treat group as other in the Windows case. If it makes it easier, feel tree to change the API to take an int rather than a `Set<PosixFilePermission>` so that you can easily check the bits.

rahulmutt:
@indegro Will you have time to finish this up? This bug fix is important for Windows users and it would great to resolve it asap.

indegro:
@rahulmutt Hello, sorry for late response. Done. Please, check and verify

indegro:
@rahulmutt I've signed CLA.

rahulmutt:
Awesome thanks! Merging now.

PR #620, opened 2018-01-24
Add IOException in the Java snippet

y-taka-23:
I'm afraid that the compilation will fail due to the checked exception.


Jyothsnasrinivas:
Thanks @y-taka-23

Issue #609, opened 2018-01-10
Trigger the linker phase when linked files other than modules change 


jneira:
Actually eta does not regenerate the main jar if the jars included as input files to be linked change. 
If you do a `etlas --make Module.hs module-dep.jar` and later `module-dep.jar` changes, a second call doesn't generate a new run-module.jar with those changes. This is so even if you add or remove jars from the call.
A full build including compile would not be neccesary, only a new linking.

jneira:
The change could be modify the [ETA.Main.DriverPipeline.linkingNeeded](https://github.com/typelead/eta/blob/master/compiler/ETA/Main/DriverPipeline.hs#L1280-L1304) to take in account all the files involved in the link phase. 
Actually [ghc](https://github.com/ghc/ghc/blob/ghc-7.10/compiler/main/DriverPipeline.hs#L435-L467) check if linking is needed:
- checking if any of object files or extra_ld_inputs are newer than the last executable
- doing a similar check for the included haskell libraries
- saving in the executable the call to linker to [compare it to next calls and trigger the link phase if they don't match](https://github.com/ghc/ghc/blob/ghc-7.10/compiler/main/DriverPipeline.hs#L471-L486).

A posiible solution for eta could be save the linked info with all files involved in the final jar (f.e. in a file called .link-info in the root). 
This info could be a line for each file linked with the name and a hash. In each call to eta, it would build that info and compare it with the last one, triggering the link if they don't match.
The hash of each file could be a checksum of the file content or a hashed string with the full path, modification time and size.

As `__extras.jar` with the java sources is one of those files we could use its hash to trigger the compile and link of java sources when they change, as required in the issue #594.

rahulmutt:
I like this solution a lot better and as you said, it'll fix a lot of problems at once.

jneira:
Closed by #622

Issue #594, opened 2017-12-19
[etlas] Recompilation checking for java source files


rahulmutt:
If you change the Java sources listed in the `java-sources:` field, `etlas` will not rebuild the new file since it doesn't do any checking for that at the moment. It should be changed so that it does.

jneira:
The change could be compute the input files hashes, saving them and compare the possible new input files in subsequent calls to trigger either a compile or only a linking if modules had not been changed and is possible.
This can be done in etlas at first but at some point the compiler could be changed so it will have this behaviour.

jneira:
* With #622 eta makes a new jar if java sources change

Issue #582, opened 2017-12-12
[etlas] Incorrect classpath to package library in the package executable launch script


jneira:
When you install a package which has a library and a executable, the launch script of the latter seems to have a incorrect path for its lib when you executes the global copy in `$ETLAS_DIR/bin` (f.e. with `etlas exec exec-name` instead `etlas run`)
The nix script generated contains:
```shell
$ETA_JAVA_CMD $JAVA_ARGS $JAVA_OPTS $ETA_JAVA_ARGS -classpath "$DIR\wai-servlet-jetty-example.jar;$DIR/../HSwai-servlet-handler-jetty-0.1.0.0-DMCQsc8EeUMHtnbzY4tt1c.jar;other_deps;$ETA_CLASSPATH" eta.main "$@"
```
The package lib jar (`HSwai-servlet-handler-jetty-0.1.0.0-DMCQsc8EeUMHtnbzY4tt1c.jar`) is not there for the script copied in $ETLAS_DIR/bin (although it works for the script placed in `$project/dist/...` )
Same for windows in the MANIFEST.MF of the launcher.jar

jneira:
I think we could change the path of the lib to point to the etlas database in the nix script or windows launcher.jar in the dist dir of the project to copy that version to etlas/bin, when doing `etlas install` . 
Cause each `etlas build` or `install` regenerates de dist dir it would point alway to the correct version of the lib (the project local copy with `build` and the global lib with `install`).
Do you know what data in the LocalBuildInfo or ComponentLocalBuildInfo tells us if the command is build or install?

rahulmutt:
At what point in the execution do you need to determine whether it's a build or install?

jneira:
I think the ideal point would be the [`libPaths` subexpression of `getDependencyClassPaths` function](https://github.com/typelead/etlas/blob/master/etlas-cabal/Distribution/Simple/Eta.hs#L638-L645) in Eta.hs

jneira:
As commented with @rahulmutt, i am exploring the possible solution of generating a special script/jar launcher in the install phase with the correct classpath an copy them to ~/etlas/bin instead of the launchers generated in the build phase.
The current wip version is in the [fix-global-exec branch of my etlas fork](https://github.com/typelead/etlas/compare/master...jneira:fix-global-exec?expand=1)

jneira:
Finally i got a initial working version that handles the three possibles types of library dir
* relative: used by `Distibution.Simple.ETA.buildExe` 
[...]

Issue #285, opened 2017-04-11
[etlas] configure --enable/disable-uberjar-mode should clear the dist directory


rahulmutt:

jneira:
@rahulmutt  Maybe this one os not longer needed due to #622 and `etlas new-build` by default

rahulmutt:
@jneira Yes, you're right! This issue is fixed.

PR #622, opened 2018-01-24
Handling relinking using persistent link info

jneira:
As commented in #609 we need to trigger the linker phase of eta if some of the values affecting the link phase change. In this way we could get eta to do a relinking when switching between uberjar and normal mode, as required in #582. 
Following loosely ghc behaviour the linker phase data is stored in the executable as a file named `.eta-link-info` in the main jar built.
The info collected is:
* The compression method: changing it requires a relink
* The paths to all files involved:
  * The main class and manifest: they are fixed but changing between lib and executable modes will trigger a relink
  * The object jars with the compiled eta modules
  * The jars of packages on which it depends if they should be included
  * The jars informed as linkable objects and the `__extras.jar` with the java and class files ones

* All the info is hashed to avoid leaks of info about the file system:
  * For most files the hash is done with the file path: compute the hash with the content of all of them could be expensive; it means the link will be trigged on file path changes and no on real content ones.
  * But this version is using the hash for the content for `__extras.jar`, with the class and java files, so we get a relinking when those files change, as required in #594
* Before comparing the last link info stored in the exe jar file and the new one, eta was checking if the linkables or packages jars had been modified after the last exe jar file. This version has added the input jars in that check but no `__extras.jar` cause it is always generated.

Maybe those changes in the linker phase makes unnecessary cleaning the dist dir when change between uberjar and normal mode (#285)

### DIFF
--- a/compiler/ETA/Main/DriverPipeline.hs
+++ b/compiler/ETA/Main/DriverPipeline.hs
@@ -80,7 +80,7 @@ import ETA.TypeCheck.TcRnTypes
 import ETA.Main.Hooks
 
 import ETA.Utils.Exception
-import ETA.Utils.Fingerprint
+-- import ETA.Utils.Fingerprint
 -- import qualified ETA.Utils.Exception as Exception
 -- import Data.IORef       ( readIORef )
 import System.Directory
@@ -89,9 +89,9 @@ import System.FilePath
 import System.PosixCompat.Files (fileExist, touchFile)
 import Control.Monad hiding (void)
 import Data.Foldable    (fold)
-import Data.List        ( partition, nub )
-import Data.Map (Map)
-import qualified Data.Map as Map
+import Data.List        ( partition, nub, union )
+-- import Data.Map (Map)
+-- import qualified Data.Map as Map
 import Data.Maybe
 -- import System.Environment
 -- import Data.Char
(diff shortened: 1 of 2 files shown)


compiler/ETA/Main/DriverPipeline.hs
```diff
@@ -1283,25 +1298,38 @@ linkingNeeded dflags linkables pkgDeps = do
...
-          if any isNothing pkgLibFiles then
-            return True
+          if any isNothing pkgLibFiles then do
+             debugTraceMsg dflags 3 (text $ "linkingNeeded: any isNothing pkgLibFiles")
+             return True
           else do
-            eLibTimes <- mapM (tryIO . getModificationUTCTime)
-                              (catMaybes pkgLibFiles)
+            extras <- getExtrasFileName dflags
+            let -- Currently __extras.jar is always generated even when modules are not compiled
```
rahulmutt: Is this something that should be changed in the future? If so please file an issue so we can track it - thanks!

jneira: Mmm the change could be hash the java and class files instead the final `__extras.jar` and avoid its own creation if the original files dont change. But now the eta packages usually have few java sources files and i think we would not reduce the build time a lot

rahulmutt: Agreed. Thanks for the explanation!

compiler/ETA/Main/DriverPipeline.hs
```diff
@@ -1343,38 +1422,48 @@ linkGeneric dflags oFiles depPackages = do
...
     --          ++ " these options."))
     -- TODO: Use conduits to combine the jars
+    linkInfoFile <- getLinkInfoFile dflags oFiles depPackages
     mainFiles' <- maybeMainAndManifest dflags isExecutable
-    mainFiles <- forM mainFiles' $ \(a, b) -> do
+    mainFiles <- forM (linkInfoFile : mainFiles') $ \(a, b) -> do
                    a' <- mkPath a
                    return (a', b)
     outJars <- mapM getNonManifestEntries oFiles
+    pkgLibJars <- getPackageLibJars dflags depPackages
```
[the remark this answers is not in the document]
jneira: Right, i'll move the check to guard `getPackageLibJars` itself in both cases

jneira:
Mmm i think i did a unwanted merge:  15cf674, i should do a reset/rebase/forced push?

rahulmutt:
Just realized that there are merge conflicts. From your forked branch, do these steps:

```
git reset --hard HEAD~1
git pull origin master --rebase
git push -f
```

jneira:
I think i get to clean up the mess with git, thanks!

Issue #604, opened 2018-01-06
Put all the mem(cpy, cmp, set) utility functions in MemoryManager


rahulmutt:
When doing patches, we find ourselves rewriting the same functions over and over. These common utility functions should be put inside the MemoryManager.

jneira:
Hi, reviewing eta i've found in libraries/base/java-utils/Utils.java the function [c_memcpy](https://github.com/typelead/eta/blob/master/libraries/base/java-utils/Utils.java#L267-L273): maybe those function could be overlodaded version of this one, for example:
```java
   // This already exists in Utils.java
   public static long c_memcpy(long destAddress, long srcAddress, int size) {
        ByteBuffer src  = MemoryManager.getBoundedBuffer(srcAddress);
        ByteBuffer dest = MemoryManager.getBoundedBuffer(destAddress);
        src.limit(src.position() + size);
        dest.put(src);
        return destAddress;
    }
   // New ones
   public static ByteBuffer c_memcpy(ByteBuffer destBuffer, ByteBuffer srcBuffer, int size) {
        ByteBuffer dest = destBuffer.duplicate();
        ByteBuffer src = srcBuffer.duplicate();
        int srcOffset = src.position();
        src.limit(srcOffset + size);
        dest.put(srcBuffer);
        return dest;
    }

    public static ByteBuffer c_memcpy(StgByteArray destArray, StgByteArray srcArray, int size) {
        return c_memcpy(destArray.buf, srcArray.buf, size);
    }

    public static ByteBuffer c_memcpy(StgByteArray destArray, ByteBuffer srcBuffer, int size) {

        return c_memcpy(destArray.buf, srcBuffer, size);
    }

    public static ByteBuffer c_memcpy(ByteBuffer destBuffer, StgByteArray srcArray, int size) {
        return c_memcpy(destBuffer, srcArray.buf, size);
    }
```

* same for cmp and set

jneira:
Hi, i have a [initial version](https://github.com/typelead/eta/compare/master...jneira:mem-utils?expand=1) with most of the methods seen in eta-hackage patches.
I am reviewing eta-hackage to identificate more generic methods over Memory and ByteBuffers

rahulmutt:
#629 resolved this, closing.

PR #629, opened 2018-01-31
Add util methods related with eta memory management

jneira:
To achieve #604


rahulmutt:
@jneira Looks great - can you shift the methods you put in base into the MemoryManager instead?

jneira:
> @jneira Looks great - can you shift the methods you put in base into the MemoryManager instead? 
yeah, i was not sure where was the best place

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +238,154 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+    }
+    
+    public static ByteBuffer copyByteBuffer( ByteBuffer src, ByteBuffer dest, int n) {
+        return copyByteBuffer(src, 0, dest, 0, n);
+    }
+    
+    public static ByteBuffer copyByteBuffer( ByteBuffer src, int srcOffset
+                                           , ByteBuffer dest, int destOffset
+                                           , int n) {
+        dest = dest.duplicate();
```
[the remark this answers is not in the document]
rahulmutt: We can make the convention this: the caller is expected to duplicate the buffer if they don't want the buffer's state mutated.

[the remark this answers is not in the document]
jneira: I remove the duplicates from the default method and add one with its own name

libraries/base/java-utils/Utils.java
```diff
@@ -40,6 +40,7 @@
 import eta.runtime.stg.StgContext;
 import eta.runtime.stg.Closure;
 import eta.runtime.io.MemoryManager;
+import eta.runtime.io.ByteArray;
```
rahulmutt: This import is no longer required right?

jneira: right

jneira:
To avoid break existing uses of MemoryManager i am writing some tests on several of its methods, in [this branch](https://github.com/jneira/eta/tree/mem-test).

jneira:
I've implemented tests on several methods of MemoryManager
Some of them ara `Ignore`d, i'll add comments for them

rts/src/test/java/eta/runtime/io/MemoryManagerTest.java
```diff
@@ -0,0 +1,295 @@
...
+                   address > 0);
+        }
+    } 
+
+    @Test(expected = IllegalArgumentException.class)
+    public void allocateBufferWithNegativeArg() {
+        allocateBuffer(-1,true);
+    }
+    @Test
+    @Ignore
```
jneira: This one throws a StackOverflowError, it is tracked in #639

rts/src/test/java/eta/runtime/io/MemoryManagerTest.java
```diff
@@ -0,0 +1,295 @@
...
+        ByteBuffer b2 = getBoundedBuffer(addr);
+        assertThat("Two invocations with the same adress ,"+
+                   "should return the same buffer", b2, is(b));
+        b2 = getBoundedBuffer(addr,0,b.remaining());
+        assertThat("Calling it with the same address, "+ 
+                   "without offset and the same length ,"+
+                   "should return the same buffer", b2, is(b));
+    }
+
+    @Ignore
```
jneira: In MemoryManager.getBoundedBuffer there is a comment that says: 
> Throws an exception  if the block that corresponds to the address has been freed
But this test doesn't throw an exception, we'd should fix the code or the comment

rahulmutt: Thanks for catching this. I've filed https://github.com/typelead/eta/issues/644 to track it. It's harmless for now but it can make debugging significantly easier if implemented.

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+    }
+
+    public static ByteBuffer copyByteBufferDups( ByteBuffer src, ByteBuffer dest, int n) {
+        return copyByteBufferDups(src, 0, dest, 0, n);
+    }
+    
+    public static ByteBuffer copyByteBuffer( ByteBuffer src, int srcOffset
+                                           , ByteBuffer dest, int destOffset
+                                           , int n) {
+        if (src == dest)
```
rahulmutt: Will this yield the correct behavior when `srcOffset != destOffset`?

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+        ByteBuffer src  = getBoundedBuffer(srcAddress);
+        ByteBuffer dest = getBoundedBuffer(destAddress);
+        return copyByteBuffer(src,srcOffset,dest,destOffset,size);
+    }
+
+    public static ByteBuffer copy(long srcAddress, long destAddress, int size) {
+        return copy(srcAddress, 0, destAddress, 0, size);
+    }
+
+    public static ByteBuffer set(long address, int val, int size) {
```
rahulmutt: Maybe this API should take a `byte val` instead of `int val` to clearly convey that only 8 bits will be written? The callee can do the casting.

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+        return buf;
+    }
+
+    public static long allocateAndSet(byte[] bytes) {
+        long address = allocateBuffer(bytes.length,true);
+        set(address,bytes);
+        return address;
+    }
+    
+    public static ByteBuffer move(long srcAddress, long destAddress, int size) {
```
rahulmutt: Ok, I see you've put src and dest, consistent with the ordering elsewhere. I think when calling this in `c_memmove`, you swapped the arguments by mistake.

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+    public static int compare(long a1, int o1, ByteBuffer b2, int n) {
+        return compare(getBoundedBuffer(a1,o1,n), b2, n);
+    }
+
+    public static int compare(ByteBuffer b1, long a2, int o2, int n) {
+        return compare(b1, getBoundedBuffer(a2,o2,n), n);
+    }
+
+    public static ByteBuffer chr(ByteBuffer b, int c, int n) {
+        c = (int)((byte) c);
```
rahulmutt: Same comment as above - maybe the `int c` shoud be `byte c` to convey that only the low 8 bits will be used?

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+        b = b.duplicate();
+        int idx = chrIndex(b, c, n);
+        if (idx == -1)
+            return emptyBuffer;
+        else
+            return (ByteBuffer) b.position(b.position() - 1);
+    }
+
+    public static int chrIndex(ByteBuffer b, int c, int n) {
+        c = (int)((byte) c);
```
rahulmutt: Same comment as above - maybe the `int c` shoud be `byte c` to convey that only the low 8 bits will be used?

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+        for (int i = 0; i < n ; i++) {
+            byte nxt = b.get();
+            if (nxt == c)
```
rahulmutt: `c` should probably be declared as `byte` since only comparisons to another `byte`'s being done.

rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -237,4 +242,161 @@ public static long loadStringsUTF8(String[] ss) throws UnsupportedEncodingExcept
...
+              return i;
+        }
+        return -1;
+    }
+    // from 
+    public static int chrOffset(long address, int startofs, int endofs, int c) {
```
rahulmutt: Same comment as above - maybe the int c shoud be byte c to convey that only the low 8 bits will be used?

rts/src/main/java/eta/runtime/io/ByteArray.java
```diff
@@ -41,39 +41,81 @@ private ByteArray(long bufferAddress, int size) {
...
+
+    public static ByteBuffer copyByteArray(ByteBuffer srcBuffer, ByteArray destArray, int size) {
+        return copyByteArray(srcBuffer, 0, destArray,0, size);
+    }
+
+    public static ByteBuffer copyByteArray(ByteArray srcArray, ByteBuffer destBuffer, int size) {
+        return copyByteArray(srcArray, 0, destBuffer, 0, size);
+    }
+
+    public static int compare(ByteArray b1, ByteArray b2, int n) {
```
rahulmutt: From this line (103) and below, all the uses of `duplicate()` are unnecessary since the ByteBuffers are created as intermediates in all the cases and won't leak outside these methods.

rts/src/main/java/eta/runtime/io/MemoryManager.java
jneira: There is [a test for that case](https://github.com/typelead/eta/pull/629/files/491c564a4a756fb29b9c430c567e49669d198e43#diff-3a696de3cc773184d32d7b592119ff16R39), although maybe it is not correct...

rts/src/main/java/eta/runtime/io/MemoryManager.java
jneira: yeah, it is cleaner, i think i copied some of the implementations of eta-hackage

rts/src/main/java/eta/runtime/io/MemoryManager.java
jneira: yeah, thanks for catching it

jneira:
Thanks for the careful review, i am afraid i've not checked all the files enough

PR #638, opened 2018-02-04
Update subclass tutorial example so it compiles

andrewgmcgregor:


andrewgmcgregor:
I ran ./install.sh and it seemed to pass. 
Are those failing tests likely to be because of my commit? If so how should reproduce them on my machine?
Cheers

Jyothsnasrinivas:
Thanks @andrewgmcgregor!  Don't worry about the failing tests. Your changes don't affect them.

PR #654, opened 2018-02-14
Add 0 to the end of allocated strings

jneira:
To mark always its termination. It fixes https://github.com/jneira/wai-servlet/issues/16

### DIFF
--- a/rts/src/main/java/eta/runtime/io/MemoryManager.java
+++ b/rts/src/main/java/eta/runtime/io/MemoryManager.java
@@ -195,9 +195,10 @@ public class MemoryManager {
         Long address = loadedStrings.get(s);
         if (address == null) {
             byte[] bytes    = s.getBytes(charset);
-            address         = allocateBuffer(bytes.length, false);
+            address         = allocateBuffer(bytes.length + 1, false);
             ByteBuffer dest = getBoundedBuffer(address);
             dest.put(bytes);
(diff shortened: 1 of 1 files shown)


rts/src/main/java/eta/runtime/io/MemoryManager.java
```diff
@@ -210,13 +211,14 @@ public static long loadStrings(String[] ss, String charset) throws UnsupportedEn
...
                 byte[] ssBytes = ss[i].getBytes(charset);
-                size += ssBytes.length;
+                size += ssBytes.length + 1;
                 bytess[i] = ssBytes;
             }
             address         = allocateBuffer(size, false);
             ByteBuffer dest = getBoundedBuffer(address);
             for (int i = 0; i < ss.length; i++) {
                 dest.put(bytess[i]);
+                dest.put((byte) 0);
```
rahulmutt: In the multi-string case, I think you should increment the size by 1 *outside* the loop since we only need a single 0 byte at the end.

`loadStrings` handles the case where you have a literal string that is larger than 65,535 bytes (and hence can't be stored in the constant pool), and so you split it up into smaller pieces and assemble them at runtime into a giant string.

jneira: oh, i did not understand its intention

Issue #372, opened 2017-05-12
Rename `pureJava` and `pureJavaWith` to `unsafePureJava` and `unsafePureJavaWith`


rahulmutt:

aiya000:
I think so too strongly 👍

PR #678, opened 2018-02-24
Fixing #372: Renaming pureJava and pureJavaWith with unsafePerformJav…

mbrc12:
…a and unsafePerformJavaWith


rahulmutt:
@mbrc12 Much cleaner, thanks! Let's see what the eta-hackage build says.

Issue #662, opened 2018-02-17
Expose runtime configuration as properties


rahulmutt:
The runtime system is quite configurable
https://github.com/typelead/eta/blob/master/rts/src/main/java/eta/runtime/Runtime.java#L16-L158

But we need a better interface. We should read from properties to initialize those values, so that we can do something like

```
java -Deta.rts.maxLocalSparks=50 ...
```

to configure the runtime.

Please list out the property names that you will be selecting before continuing further.

ouromoros:
As I see it, using properties would involve reading a property fiile(probably `~/.etlas/.properties`?). So I think adding a initialization step before running the program would do it.

Snce `getPropertie` returns only String, we would have to parse the return value to the corresponding type for every paramater. I think a helper function is needed for this parsing task or there would be *many* tedious try-catch phrases and the code won't be very extencible.

I think all the paramaters in the referenced lines can be added, or is there something I'm missing? I'll work on it if you think it's fine.

rahulmutt:
The properties file should actually be a resource in the compiled JAR file:
https://www.javaworld.com/article/2077352/java-se/smartly-load-your-properties.html

This is because you would want to configure the Eta runtime in an application-specific manner and not the same way for every application.

We can make a convention that we expect it to be `eta/rts.properties`. So there are two levels of properties here:
1) Properties from the `eta/rts.properties` file that has been added as a *resource*.
2) System properties provided on the command line via `-D` flag.

For a given property, the value specified by 2) should override the value specified by 1).

Go ahead and work on this :) 

I agree that the bulk of the work will be parsing the options. You can create a new class `RuntimeOptions` and put all your parsing code there.

rahulmutt:
@ouromoros Before you waste your time writing parsing utilities, you can use the following methods:
- `Boolean.getBoolean("[property-name]")` for parsing boolean properties,
- `Integer.getInteger("[property-name]")` for parsing integer properties

and so on for all the standard primitive wrapper classes.

rahulmutt:
Disregard my last comment. You still have to parse the properties that are read from the file.

rahulmutt:
Resolves with #682.

PR #682, opened 2018-02-28
Expose various parameters to be configuable via Java properties.

ouromoros:
##### To achieve https://github.com/typelead/eta/issues/662
I'm wondering if I should write test for it.(shamefully I would have to learn using Junit if so)

Also, I'm not sure whether a switch statement using Strings or some if-else checking directly passed class type is better, as in `getValue` method. Though generally it shouldn't hurt much.

### DIFF
--- a/rts/src/main/java/eta/runtime/Runtime.java
+++ b/rts/src/main/java/eta/runtime/Runtime.java
@@ -15,14 +15,15 @@ import eta.runtime.exception.RuntimeInternalError;
 
 public class Runtime {
 
-    /** Runtime Parameters **/
+    /**
+     * Runtime Parameters
+     **/
 
     public static final String RTS_PROPERTIES_PATH = "eta/rts.properties";
 
     /* Parameter: maxWorkerCapabilities (int)
        The total number of Capabilities that can be spawned by the runtime itself. */
-    private static int maxWorkerCapabilities
-        = getNumberOfProcessors();
+    private static int maxWorkerCapabilities;
 
     public static int getMaxWorkerCapabilities() {
         return maxWorkerCapabilities;
@@ -34,7 +35,7 @@ public class Runtime {
 
     /* Parameter: maxGlobalSparks (int)
        The total number of sparks that are allowed in the Global Spark Pool at a time. */
-    private static int maxGlobalSparks = 4096;
+    private static int maxGlobalSparks;
 
     public static int getMaxGlobalSparks() {
         return maxGlobalSparks;
@@ -132,7 +133,7 @@ public class Runtime {
 
     public static boolean setDebugMode(char c) {
         boolean valid = true;
-        switch(c) {
+        switch (c) {
             case 's':
                 debugScheduler = true;
                 break;
@@ -161,11 +162,16 @@ public class Runtime {
         return debugMemoryManager;
     }
 
-    public static void init_params(){
+    public static void init_params() {
         RuntimeOptions rto = new RuntimeOptions(RTS_PROPERTIES_PATH);
-        /*
-         * TODO: Initialize parameters explicitly
-         */
+        // Initialize parameters explicitly
+        maxWorkerCapabilities = rto.getInt("maxWorkerCapabilities", getNumberOfProcessors());
+        maxGlobalSparks = rto.getInt("maxGlobalSparks", 4096);
+        minTSOIdleTime = rto.getInt("minTSOIdleTime", 20);
+        maxTSOBlockTime = rto.getInt("maxTSOBlockTime", 1);
+        minWorkerCapabilityIdleTime = rto.getInt("minWorkerCapabilityIdleTime", 1000);
+        gcOnWeakPtrFinalization = rto.getBoolean("gcOnWeakPtrFinalization", false);
+        maxLocalSparks = rto.getInt("maxLocalSparks", 4096);
     }
 
     public static void main(String[] args, Closure mainClosure) throws Exception {
@@ -228,14 +234,17 @@ public class Runtime {
     public static void maybeFlushStdHandles() {
         try {
             evalIO(Closures.flushStdHandles);
-        } catch (Exception e) {}
+        } catch (Exception e) {
+        }
     }
 
     public static void stgExit(int code) {
         System.exit(code);
     }
 
-    /** Command Line Arguments **/
+    /**
+     * Command Line Arguments
+     **/
 
     private static final String RUNTIME_NAMESPACE = "eta.runtime";
     private static String[] programArguments;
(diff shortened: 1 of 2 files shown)


ouromoros:
Actually I think I'm going to re-implement it in an OOP style. So if anyone sees this, maybe review later...

rts/src/main/java/eta/runtime/RuntimeOptions.java
```diff
@@ -0,0 +1,47 @@
...
+        Properties p = new Properties();
+        // Load properties from resource
+        try(InputStream in = RuntimeOptions.class.getClassLoader().getResourceAsStream(RTS_PROPERTIES_PATH)){
+            if(in != null) p.load(in);
+        }catch(IOException e){
+            e.printStackTrace();
+        }
+        Properties sp = System.getProperties();
+        // Set loaded properties to be override by System properties
+        for(String key : sp.stringPropertyNames()){
```
rahulmutt: You're looping through *all* the System properties with `stringPropertyNames()`. Instead, loop over just the static list of Eta runtime properties.

rahulmutt:
Great progress so far @ouromoros. Keep at it!

ouromoros:
I have finished. Here's what I've done so far:
1. Implement `RuntimeOptions` class to access values provided by java properties.
2. Initialize all the parameters(except debug) in `init_params` method and delete the old initialization code.

rts/src/main/java/eta/runtime/Runtime.java
```diff
@@ -157,8 +162,21 @@ public static boolean debugMemoryManager() {
...
+        RuntimeOptions rto = new RuntimeOptions(RTS_PROPERTIES_PATH);
+        // Initialize parameters explicitly
+        maxWorkerCapabilities = rto.getInt("maxWorkerCapabilities", getNumberOfProcessors());
+        maxGlobalSparks = rto.getInt("maxGlobalSparks", 4096);
+        minTSOIdleTime = rto.getInt("minTSOIdleTime", 20);
+        maxTSOBlockTime = rto.getInt("maxTSOBlockTime", 1);
+        minWorkerCapabilityIdleTime = rto.getInt("minWorkerCapabilityIdleTime", 1000);
+        gcOnWeakPtrFinalization = rto.getBoolean("gcOnWeakPtrFinalization", false);
+        maxLocalSparks = rto.getInt("maxLocalSparks", 4096);
+    }
```
rahulmutt: All the keys should be prefixed with `eta.rts.` - this is a change you can do inside of `getValue` probably. 

Moreover, please define static constants for all the keys (i.e. `public static final MAX_WORKER_CAPABILITIES_PROPERTY = "maxWorkerCapabilities";`, etc.)

rts/src/main/java/eta/runtime/Runtime.java
```diff
@@ -157,8 +162,21 @@ public static boolean debugMemoryManager() {
         return debugMemoryManager;
     }
 
+    public static void init_params() {
```
rahulmutt: Please name this method `initializeRuntimeParameters`.

rts/src/main/java/eta/runtime/Runtime.java
```diff
@@ -157,8 +162,21 @@ public static boolean debugMemoryManager() {
...
+        minTSOIdleTime = rto.getInt("minTSOIdleTime", 20);
+        maxTSOBlockTime = rto.getInt("maxTSOBlockTime", 1);
+        minWorkerCapabilityIdleTime = rto.getInt("minWorkerCapabilityIdleTime", 1000);
+        gcOnWeakPtrFinalization = rto.getBoolean("gcOnWeakPtrFinalization", false);
+        maxLocalSparks = rto.getInt("maxLocalSparks", 4096);
+    }
+
     public static void main(String[] args, Closure mainClosure) throws Exception {
         Runtime.setProgramArguments(args);
+        init_params();
```
rahulmutt: Foreign exports (calling Eta from Java) doesn't always call `Runtime.main` so the parameters won't get initialized in that case. Instead, place the call to parameter initialization in a static initializer for this class.

rts/src/main/java/eta/runtime/RuntimeOptions.java
```diff
@@ -0,0 +1,74 @@
...
+        Properties lp = new Properties();
+        // Load properties from resource
+        try (InputStream in = getClass().getClassLoader().getResourceAsStream(path)) {
+            if (in != null) lp.load(in);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        Properties sp = System.getProperties();
+        // Set loaded properties to be override by System properties
+        for (String key : lp.stringPropertyNames()) {
```
rahulmutt: `lp.stringPropertyNames()` will be empty if there's no properties file resource, so this will not override with System properties which is the expected behavior.

rts/src/main/java/eta/runtime/RuntimeOptions.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public Properties getProperties() {
+        return p;
+    }
+
+    /*
+     * Get and parse value from properties.
+     * Return default value if failed.
+     */
+    private Object getValue(String key, String type, Object d) {
```
rahulmutt: Rather than using a generic `getValue` method, why not put the specific code inside the respective `getInt`, `getBoolean`, `getDouble` functions? Yes, there will be repetition but I think it'll be a bit clearer and avoids unnecessary boxing for primitive types.

rts/src/main/java/eta/runtime/RuntimeOptions.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public RuntimeOptions(String path) {
+        load_properties(path);
+    }
+
+    /*
+     * Loads properties from resource which
+     * may be override by system properties
+     */
+    private void load_properties(String path) {
```
rahulmutt: Please name this as `loadProperties` according to the naming convention of the rest of the source code.

rts/src/main/java/eta/runtime/RuntimeOptions.java
ouromoros: Actually this is a comment I forgot to change here. What the code actually do is add properties loaded from resource to system poperties without overriding existing ones. Then use `sp` as the actual properties we read from.

rahulmutt: My bad, you're right! Thanks for the clarification.

ouromoros:
@rahulmutt Thanks for the suggestions! I have done the requested changes now.
Regarding the keys for retrieving properties, I think we'd better keep them out of RuntimeOptions class to reduce coupling.

rahulmutt:
@ouromoros Awesome work! Looks perfect. Is there anything else you wanted to add or should I merge?

ouromoros:
@rahulmutt Thanks! I think that'll be all. Please merge.

Issue #689, opened 2018-03-04
createDirectoryIfMissing doesn't work (throws unexpected error)


kemphack:
`jCreateDirectoryIfMissing dir = io $ createDirectoryIfMissing True (fromJava dir)`

`foreign export java "@static Tools.createDirectoryIfMissing"
    jCreateDirectoryIfMissing :: JString -> Java a ()`

`Tools.createDirectoryIfMissing("./folder");`
does work but 
`Tools.createDirectoryIfMissing("./folder/folder");` 
doesn't: throws error
`Exception in thread "main" eta.runtime.exception.EtaException: JException java.nio.file.NoSuchFileException: folder/folder`

NickSeagull:
Is this the function from `System.Directory`? If so, you should pass `True` as the first parameter so it creates parents. If not, it will throw.

kemphack:
` io $ createDirectoryIfMissing True (fromJava dir)` it throws with True. Yes, it's from System.Directory

NickSeagull:
Could you please add the bytecode and the STG info as described [here](https://github.com/typelead/eta/blob/master/docs/source/eta-user-guide.rst#debugging-stack-traces)? Thanks! 😄

kemphack:
I don't think that bytecode of IOError can be usefull in this case. Something bad happens inside of java realization of createDirectoryIfMissing:
`eta.runtime.exception.EtaException: JException java.nio.file.NoSuchFileException: folder/folder/folder
	at base.system.io.Error.tryIOError2(Error.hs)
	at base.system.io.Error$tryIOError2.apply1V(Error.hs)
	at eta.runtime.exception.Exception.catch_(Exception.java:166)
	at directory.system.Directory.$Lr1CS8a2(Directory.hs:350)
	at directory.system.Directory.$Lr1CSAa3(Directory.hs:346)
	at directory.system.Directory.createDirectoryIfMissing1(Directory.hs:341)
`

NickSeagull:
🤔 Can you create a repo with this issue to see if I can try to reproduce?

kemphack:
Here it is https://github.com/kemphack/directoryMissingBug

jneira:
hi! i've check the code of directory and[ it catches an `IOError` to check if it have to create the intermediate folders (sic!)](https://github.com/haskell/directory/blob/master/System/Directory.hs#L335)
As java is throwing a JException instead an IOError the check fails.
[...]

PR #700, opened 2018-03-12
Map java exceptions to haskell IOExceptions (aka IOError)

jneira:
Hi, analyzing [an issue](https://github.com/typelead/eta/issues/689) involving [tryIOError](https://github.com/typelead/eta/blob/master/libraries/base/System/IO/Error.hs#L102) we detected that java exceptions involving file io errors was not catched (file not found, file already exists, etc).
Talking with @rahulmutt, he suggested that maybe we could do a mapping between the appropiate java exceptions and haskell IOErrors to make System.IO.Error catching work.
This pr is a wip around that idea, we put the mapping in the Exception instance for JException to make it global. So we should study carefully the implications of that change.

Examples: 
* This code behaves as expected:
```haskell
{-# LANGUAGE MagicHash,TypeFamilies,DataKinds,
    FlexibleContexts,OverloadedStrings, ScopedTypeVariables #-}
module Main where
import Java
import Java.Exception
import qualified System.IO.Error as SysIOErr
import Data.Proxy
import Data.Typeable
import qualified Control.Exception as EX 

foreign import java "@new" newFileNotFoundException ::  String -> FileNotFoundException
foreign import java "@new" newException ::  String -> JException

main = do
  let ioEx = newFileNotFoundException "io ex"
      ex   = newException "ex"
  r <- SysIOErr.tryIOError (EX.throwIO ioEx)
  case r of
    Right() -> return ()
    Left err -> putStrLn $ EX.displayException err
  (r' :: Either JException ()) <- EX.try (EX.throwIO ex)
  case r' of
    Right() -> return ()
    Left err -> putStrLn $ EX.displayException err
  (r' :: Either EX.IOException ()) <- EX.try (EX.throwIO ioEx)
  case r' of
    Right() -> return ()
    Left err -> putStrLn $ EX.displayException err
```
output:
```
does not exist (io ex)
JException java.lang.Exception: ex
does not exist (io ex)
```
--------------------------
* This one seems to be correct too
```haskell
r <- SysIOErr.tryIOError (EX.throwIO ex)
  case r of
    Right() -> return ()
    Left err -> putStrLn $ EX.displayException err
```
```
Exception in thread "main" eta.runtime.exception.EtaException: JException java.l
ang.Exception: ex
....
```
* As that JException is not an IOError the exception is thrown
---------------------------
```haskell
(r' :: Either JException ()) <- EX.try (EX.throwIO ioEx)
  case r' of
    Right() -> return ()
    Left err -> putStrLn $ EX.displayException err
```
```
Exception in thread "main" eta.runtime.exception.EtaException: does not exist (i
o ex)
....
```
* As being the change global, java exceptions that are IOErrors are not being handled as JExceptions anymore. Maybe that could be a problem. If so we could move the conversion to the relevant functions of `System.IO.Error`
   * If hakell exceptions would have the concept of causing exceptions we could put the JException as cause of the IOError
---------------------
* Caveats
  * Not all types of IOErrors are covered:
    * `IllegalOperationType`
  * Dont know if all possible java exceptions related with files being used are covered
  * The code is, in general, fragile and too much tied to the wild error handling of java
  * The mapping is "closed", only is defined for the specific java exceptions. If not the default `SomeException` is created (although the java exception is a java.io.IOException)


jneira:
In the code i had to add the `a <: Throwable` constraint to `toIOError` (and `toException`) to be able to use `getMessage :: (a <: Throwable) => Java a String`. Being already the `a <: JException` constraint, ¿should be `Throwable` unnecesary?

libraries/base/Java/Exception.hs
```diff
@@ -119,11 +122,86 @@ data {-# CLASS "java.io.IOException" #-} IOException = IOException (Object# IOEx
...
+
+type instance Inherits FileAlreadyExistsException = '[FileSystemException]
+
+data {-# CLASS "java.nio.channels.OverlappingFileLockException" #-} OverlappingFileLockException =
+  OverlappingFileLockException (Object# OverlappingFileLockException)
+  deriving (Class, Show, Typeable)
+
+type instance Inherits OverlappingFileLockException = '[IllegalStateException]
+
+toIOError :: (ioex <: JException, ioex <: Throwable)
```
rahulmutt: Because `JException <: Throwable`, using `ioex <: Throwable`. Ditto below in the `Exception` instance.

jneira: Yeah, you are right, Throwable is enough.

About the place to do the conversion: it's fine to put in the `System.IO.Error`module but it would introduce some inconsistency between `tryIOError` and `System.Exception.try`. The former will catch those java exceptions as `IOError`/`IOException` but the latter will not do it, cause for `try` they will continue being `JException`s. In haskell they behave identically.
So client code will have to do the conversion "manually" if they use `try` and want to catch them as `IOError`s.
Actually `toException` converts `JException` to haskell `SomeException`, the runtime system requires that type? it could handle IOExceptions as well?

jneira: He, i didnt know i would get a class hierarchy problem in haskell ;-)

rahulmutt: @jneira It's weird, but catching `JException` will catch *any subclasses* of `java.lang.Exception` as well - if it behaved any other way, it would be very confusing for people familiar with JVM languages' exception behavior. We will have to have restricted forms of subtyping in Eta at least to make the interop smooth.

Hmm I see your point. But what if you use `try` around an FFI function that throws a `FileNotFoundException` and you try to catch it that way with the corresponding JWT - if this `IOError` conversion happens, then my catch won't work! How do we deal with thise case?

rahulmutt:
Also, to address your concern about this mapping being closed, why not make a typeclass that has `toIOError` as the typeclass method? If an typeclass instance for a JWT `<: Throwable` is provided, then the conversion will happen, otherwise it won't. (The otherwise it won't can be handled with a default implementation that returns `Nothing`).

libraries/base/Java/Exception.hs
jneira: Mmmm and what if we change the `Control.Exception.try` to make the conversion if the client code wants a IOError and it has a JException? would be it possible?
Well the change had to be in `catch#` (or `GHC.IO.catch`and `GHC.IO.catchAny`), cause all catching functions calls it. Where is defined `catch#`?

jneira:
I created the `class (e <: Throwable) => ToIOError e`, not sure if it captures your comment cause it has a default impl for all subclasses of Throwable and returns `Maybe IOError`
The problem is i was no able to make a `instance (e <: IllegalStateException) => ToIOError e` only `instance ToIOError IllegalstateException`

rahulmutt:
Hmm right, because the typeclass instance resolution requires concrete types and doesn't know how to deal with subtyping. I think we just have to accept that this is going to be a bit brittle and add new cases as we come across them. Do you have any other ideas on how to make this extensible? If not, remove the typeclass and I'll merge what you have so far. 

Have you tested it on anything yet?

jneira:
Only with the code of the first comment. 
It is pending where to do the conversion, in the Exception instance for Throwable or somewhere (see https://github.com/typelead/eta/pull/700#discussion_r174753174)
Re `ToIOError` I am trying another approach: use the type class to split the big defaultToIoError in instances for each java exception

jneira:
@rahulmutt i wonder if we can do the conversion in the runtime: `eta.runtime.exception.Exception.catch_`
if we can deduce the haskell exception type asked by the client code (from `Closure handler`?) maybe we could change `convertJavaException` to return a `IOException` instead `SomeException` if the client code want to catch a `IOException`

jneira:
In fact it seems we should do it in rts:
```haskell
 let ioEx = newFileNotFoundException "io ex"
      ex   = newException "ex"
      ioErr = mkIOError SysIOErr.doesNotExistErrorType "io err"
  
  catchException (EX.throwIO ioEx) $ \ (e :: SysIOErr.IOError) ->
     putStrLn $ EX.displayException e
  catchException (EX.throwIO ioErr) $ \ (e :: SysIOErr.IOError) ->
     putStrLn $ EX.displayException e
  catchException (readFile "x" >> return ()) $ \ (e :: SysIOErr.IOError) ->
     putStrLn $ EX.displayException e
```
The third catch doesnt work cause the root exception is `EtaException` and no `java.nio.file.NoSuchFileException`. So only left do it in rts before the wrapping or search the cause (only one level?) of the `EtaException` to find `NoSuchFileException`  to make the conversion

rahulmutt:
Oh right, forgot about the wrapping. Yes, I agree, this will most likely have to be implemented in the runtime system in the manner you described. Go ahead and try it out.

jneira:
@rahulmutt finally i get a version that works, although it is rather unambitious:
* it keeps the logic in a unique function `toIOError`
* it does not automatic conversion, only defines in `Java.Exception` two new functions to catch `JException`s as `IOError`s: `catchJavaIOError` and `tryJavaIOError`
  * client code will have to choose call them and mask JExceptions

I tried [a version changing the runtime](https://github.com/jneira/eta/blob/io-exception-rts/rts/src/main/java/eta/runtime/stg/Closures.java#L219-L230) but it will be a little bit more complex due to the interop and it would set the conversion fixed, with the problems you commented.
As far i know the type information are not in at runtime and we could not know if user code wants to catch a JException or an IOError in [`catch_`](https://github.com/jneira/eta/blob/io-exception-rts/rts/src/main/java/eta/runtime/exception/Exception.java#L123-L174) (if it is possible  let me know, as that option would be more interesting)

With the actual version the tests had been:
```haskell
import Java
import Java.Exception
import qualified System.IO.Error as SysIOErr
import qualified Control.Exception as EX 


foreign import java "@new" newFileNotFoundException ::  String -> FileNotFoundException
foreign import java "@new" newException ::  String -> JException

mkIOError type' msg = SysIOErr.ioeSetErrorString
                      (SysIOErr.mkIOError type' "" Nothing Nothing) msg

main = do
  let ioEx = newFileNotFoundException "io ex"
      ex   = newException "ex"
      ioErr = mkIOError SysIOErr.doesNotExistErrorType "ioerr"
  SysIOErr.catchIOError (EX.throwIO ioErr) $ putStrLn . show
  catchJavaIOError (EX.throwIO ioErr) $ putStrLn . show

  EX.catch (EX.throwIO ioEx) $ \ (e :: JException) -> putStrLn $ show  e
  EX.catch (EX.throwIO ex) $ \ (e :: JException) -> putStrLn $ show e

  catchJavaIOError (EX.throw ioEx) $ putStrLn . show
  catchJavaIOError (readFile "x" >> return ()) $ putStrLn . show
```
```
does not exist (ioerr)
does not exist (ioerr)
JException java.io.FileNotFoundException: io ex
JException java.lang.Exception: ex
does not exist (java.io.FileNotFoundException: io ex)
does not exist (java.nio.file.NoSuchFileException: x)
```

jneira:
I tried to change System.IO.Error.tryIOError to call tryJavaIOError and i get a compiler error:
```
Module imports form a cycle:
         module `Control.Applicative' (.\Control\Applicative.hs)
        imports `Control.Arrow' (.\Control\Arrow.hs)
  which imports `Control.Monad.Fix' (.\Control\Monad\Fix.hs)
  which imports `System.IO' (.\System\IO.hs)
  which imports `System.Posix.Internals' (.\System\Posix\Internals.hs)
  which imports `System.IO.Error' (.\System\IO\Error.hs)
  which imports `Java.Exception' (.\Java\Exception.hs)
  which imports `Data.List' (.\Data\List.hs)
  which imports `Data.Traversable' (.\Data\Traversable.hs)
  which imports `Control.Applicative' (.\Control\Applicative.hs)
```
i suppose it is a way to break the circular dependency...

jneira:
@rahulmutt Well i think we let the fix as it is now: client code have to opt in explicitly to catch IOErrors or JExceptions that can be converted to them, using `tryJavaIOError`. 
I test the changes with the `directory` package that relies heavily in catching IOErrors and it seems to work fine
If later we think that it is cumbersome for client code we could try to explore those others options described in the issue (runtime, type classes, etc)

rahulmutt:
@jneira Bumped to `0.7.1b2` https://github.com/typelead/eta/commit/faf880d85da70df0d39a378dea40a6501cdca077 as you requested.

Issue #555, opened 2017-11-15
Add conversion functions for JChar <-> Char


rahulmutt:

kpadmasola:
@rahulmutt any pointers for getting started here? Thanks!

rahulmutt:
@kpadmasola First, take a look at how `JChar` and `Char` are defined. `JChar` is a 16-bit *unsigned* integer, part of a UTF-16 encoded string, and `Char` is a 32-bit *unsigned* integer, part of a UTF-32 encoding. The conversion won't be exact, make sure the function type is `JChar -> Maybe Char`. 

`JChar` has some instances defined for it that you can make use of (it's treated as a 16-bit unsigned integer) like `Integral`, etc. You can find these in `libraries/base/Java/Primitive.hs`

alanasp:
@rahulmutt So, `Char` and `JChar` are defined similarly as `data JChar = JC# JChar#` and `data Char = C# Char#`, respectively. Since `Char#` is an embedded type, the compiler would evaluate `'x'#` expression as `Char#`. Is there a similarly defined expression, which would be interpreted as `JChar#`? I can't seem to find any docs on this and this is what the conversion of types ultimately boils down to. 
Thanks for your help.

rahulmutt:
@alanasp You fortunately don't need to deal with all the internals since it has been done for you in this module:
https://github.com/typelead/eta/blob/master/libraries/base/Java/PrimitiveBase.hs
`JChar` has an `Integral` and `Num` instance (since it's just a 16-bit signed integer) so you should be able to use any of those typeclass methods for the conversion.

Hope that helps!

NickSeagull:
What about making `JChar` and `Char` the same type, what implications could that have? :thinking:

NickSeagull:
Oh, that makes sense :smile:

alanasp:
@rahulmutt Thanks, this actually was really helpful. I have forked the stable version of eta and updated the `Java.Primitive` module to contain the conversion functions. You can see changes [here](https://github.com/alanasp/eta/commit/003d9433a77dcdefcc540e24c1ffedbe048c1189). The updated module builds and works fine on some tests that I ran. Should I move the functions to a different module or is it fine as is? If so, should I submit a pull request? Let me know what you think.

rahulmutt:
@alanasp Go ahead and submit a PR :) Every PR has a test suite run against it so we'll discover any issues there.

alanasp:
[...]

PR #712, opened 2018-03-26
Added JChar <-> Char conversion functions to Java.Primitive module

alanasp:
I have added JChar <-> Char conversion functions to Java.Primitive module. The functions check that once converted to Int the value falls within the allowed range for the other type and only then converts, otherwise returns Nothing.
#555


alanasp:
I will try to submit a pull request to Stable as it seems that the Master is failing but it is failing the build without the changes as well.

Issue #560, opened 2017-11-17
Split JavaConverter into two typeclasses to aid type inference


rahulmutt:
As @mandubian pointed out on Gitter:

```haskell
class JavaConverter a b where
  toJava :: a -> b
  fromJava :: b -> a
```
makes type inference really difficult. 

We can split it up and add functional dependencies to make it better:

```haskell
class ToJava a b | a -> b where
  toJava :: a -> b

class FromJava a b | a -> b where
  fromJava :: b -> a
```

njayinthehouse:
Hey! Can I take this one up?

Jyothsnasrinivas:
@njayinthehouse Sure! This change is a breaking change. So you have to think through how you are going to implement this properly.

njayinthehouse:
Hey! I'm working on it right now. My approach has been to:

1. Replace the class definition as shown by @rahulmutt above

2. Replace each instance declaration of JavaConverter with an instance declaration of ToJava and FromJava each.

njayinthehouse:
But just out of curiosity, how does a change like this improve type inference?

mbrc12:
I had written down the same change, only to realize that declarations like
````
instance Integral a => ToJava a    JInteger where
....
instance Integral a => ToJava a    JLong where
....
````
break FuncDeps.

**Comment by jneira** [comment]
@njayinthehouse see https://wiki.haskell.org/Functional_dependencies for a explanation on how fundeps can help to compiler to infer the type of a expression

**Comment by rahulmutt** [comment]
@mbrc12 Has a good point.

Moreover, I recently encountered a very nasty bug in the `directory` package that uses `fromJava` in its implementation and ended up performing I/O in the wrong order producing the wrong results. This is definitely not a good thing and it means we need to re-think whether the `JavaConverter` type class is a good idea in the first place.

When converting from a pure immutable Eta type, the conversion process will be side-effect free in *most* cases. I don't want to say *all* because it's Java we're talking about and there's definitely going to be an exception.

When converting a mutable Java object to an Eta type, the conversion process can have side-effects and can also fail in case there's a domain mismatch between the source and target.

So the real type signatures for `ToJava` and `FromJava` that handles all this is:

```
class ToJava a b where
  toJava :: a -> Java c (Maybe b)

class FromJava a b where
  fromJava :: b -> Java c (Maybe a)
```

Now it makes conversions harder to use, but makes sure we handle all the possible error cases that can happen when transitioning from Java to Eta and vice-versa.

[...]

PR #713, opened 2018-03-26
Added JChar <-> Char conversion functions to Java.Primitive module

alanasp:
I have added JChar <-> Char conversion functions to Java.Primitive module. The functions check that once converted to Int the value falls within the allowed range for the other type and only then converts, otherwise returns Nothing.

I have tried submitting a pull request to master but just found out that it was failing the build test on itself. This should work on stable as I have done some testing on my local machine and built it successfully.

#555

### DIFF
--- a/libraries/base/Java/Primitive.hs
+++ b/libraries/base/Java/Primitive.hs
@@ -17,7 +17,9 @@
 module Java.Primitive
   ( Byte(..)
   , Short(..)
-  , JChar(..) )
+  , JChar(..)
+  , charToJChar
+  , jcharToChar)
 where
 
 import GHC.Base
@@ -28,6 +30,9 @@ import Java.PrimitiveBase
 
 import Data.Data
 import Data.Typeable
+
+import Data.Char
+import Prelude(minBound, maxBound)
 
 deriving instance Typeable Byte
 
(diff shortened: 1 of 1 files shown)


libraries/base/Java/Primitive.hs
```diff
@@ -57,6 +62,22 @@ instance Data Short where
...
+          chrVal = ord chr
+
+jcharToChar :: JChar -> Maybe Char
+jcharToChar jchr
+    | minChar <= jchrVal && jchrVal <= maxChar = Just $ chr jchrVal
+    | otherwise = Nothing
+    where minChar = ord (minBound :: Char)
+          maxChar = ord (maxBound :: Char)
+          jchrVal = fromIntegral jchr
+
```
rahulmutt: `JChar` is a 16-bit unsigned number and `Char` is 32-bit. So the condition that it falls within the founds of `Char` will trivially hold - no need to check for bounds. What you do need to check for is whether the `JChar` is a surrogate - if it's a surrogate, then return `Nothing`.

For more details on surrogates, you can take a look [here](https://en.wikipedia.org/wiki/UTF-16).

libraries/base/Java/Primitive.hs
```diff
@@ -57,6 +62,22 @@ instance Data Short where
 
 deriving instance Typeable JChar
 
+charToJChar :: Char -> Maybe JChar
+charToJChar chr
+    | minJChar <= chrVal && chrVal <= maxJChar = Just $ fromIntegral chrVal
+    | otherwise = Nothing
+    where minJChar = fromIntegral (minBound :: JChar)
+          maxJChar = fromIntegral (maxBound :: JChar)
+          chrVal = ord chr
```
rahulmutt: You should consider surrogates (mentioned below) here too. The output `JChar` should *not* be a surrogate.

alanasp: @rahulmutt Thanks, appreciate this. Do you think that I should drop the range constraints from JChar? I thought about the fact that it holds trivially under the current definition of JChar but I added them in case it changes in the future. I will also take a look at the surrogates and submit a new PR.

jneira:
@alanasp great work! 
@rahulmutt i wonder if it worths to add an instance `JavaConverter Char JChar`, although for now it supposes use fromJust (see #560)

alanasp:
I have optimised the code to only check for bounds where necessary, given the current definition of Char and JChar. Also, the if the integer value is classed as surrogate, Nothing is returned in both functions.

rahulmutt:
@jneira That would be great. We'll have to check to make sure it doesn't have a dependency cycle though.

@alanasp You new changes look perfect. Can you create a new PR with the same changes, but against `master`?

rahulmutt:
Closing as this is redundant.

PR #721, opened 2018-04-02
Eta573

rajcspsg:


rahulmutt:
Hi @rajcspsg, it looks like you need to update your fork's `master` to the current master for `eta` and then make you changes on top of a new branch that starts from the updated `master` in yor fork.

rajcspsg:
Thanks @rahulmutt. I've updated the PR. 
Mostly I've updated the class package and refactored the usages of those classes in this PR. Only in this class https://github.com/typelead/eta/pull/721/files#diff-2f5c68896b14fe36169b53a471cd71b9  I've added getter/setters, since the default access modifiers for those fields doesn't let me to use the same in different package.

rahulmutt:
Looks like the build failed: https://circleci.com/gh/typelead/eta/1467?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Can you make sure you run `./cleaninstall.sh` with your changes and made it pass locally?

compiler/ETA/CodeGen/Closure.hs
```diff
@@ -61,20 +61,21 @@ lfStandardForm args fvs
...
   where thunkBase
-          | isTopLevel lfTopLevelFlag = "CAF"
+          | isTopLevel lfTopLevelFlag = "runtime/thunk/CAF"
           | lfUpdatable               = "UpdatableThunk"
-          | otherwise                 = "SingleEntryThunk"
+          | otherwise                 = "runtime/thunk/SingleEntryThunk"
         thunkExt
           | isTopLevel lfTopLevelFlag = mempty
-          | hasStdLayout = T.pack (show fvs)
+          | hasStdLayout = T.pack ((++) (show fvs) "runtime/thunk/")
```
rajcspsg: @rahulmutt I've tried `| hasStdLayout = T.pack ((++) (show fvs) "runtime/thunk/")` and `| hasStdLayout = T.pack ((++) "runtime/thunk/" (show fvs) )` still I get `Caused by: java.lang.ClassNotFoundException: eta.UpdatableThunk5runtime.thunk.`error. Any idea?

rahulmutt: Why are you trying to add `runtime/thunk/` instead of removing it above? Have you changes the names of all the `eta/runtime/thunk/*Thunk*` classes to `eta/*Thunk*`?

rajcspsg: @rahulmutt The classes Thunk, UpdatableThunk are only moved to package eta.
The remaining  *Thunk* classes are present in package package eta/runtime/thunk.
when I make package as eta , I get getting CNFE for java.lang.NoClassDefFoundError: eta/UpdatableThunk2.
when I make package as eta/runtime/thunk/ , I get getting CNFE for java.lang.NoClassDefFoundError: eta/runtime/thunk/UpdatableThunk.
so, I'm trying to find  way to include both eta/UpdatableThunk and eta/runtime/thunk/**.

rahulmutt: I think it will be easier if you shifted all the `eta/runtime/thunk/**` classes to `eta/*`. That way you won't have to add special cases in the logic.

rahulmutt: I think it's fine to move up the specialized thunks and functions up to the `eta` package. They will be exposed in the public runtime API for those who need it.

Issue #740, opened 2018-04-29
Iconv sets wrong values in output buffer causing partial readFile's


rahulmutt:
Anytime the `HSIConv.java` code is used for decoding large files (> 2048 bytes), `readFile` will only read fragments of the files. The reason is that the calculation done for how many bytes are remaining in the original byte buffer after an encode/decode after an Overflow `CodecResult` is not exactly correct.

To test this out: set the locale to something equivalent to ASCII so that the iconv codepath is triggered.
```
main :: IO ()
main = print =<< fmap length (readFile "myfile.txt")
```

Make sure `myfile.txt` is created with N > 2048 bytes of text. The output will not be equal to N!

The problem is this:

https://github.com/typelead/eta/blob/master/libraries/base/java-utils/HSIConv.java#L73

`inbuf` has already been completely read by:

https://github.com/typelead/eta/blob/master/libraries/base/java-utils/HSIConv.java#L131

So that the rest of the code will *think* that there is no remaining bytes to read even if there were!

@jneira You've spent a lot of time with this code - do you have any ideas how we can get an accurate calculation of how many bytes are remaining?

jneira:
Hi, your analysis is correct and i am afraid that it is difficult to compute the bytes readed in the overflow case cause we have an intermediate buffer using utf-16 between the two charsets being recoded.
I think that is inevitable to go down a abstraction level and do the decode/encode step for every char, probably using [decodeLoop](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html#decodeLoop(java.nio.ByteBuffer,%20java.nio.CharBuffer)) and [encodeLoop](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetEncoder.html#encodeLoop(java.nio.CharBuffer,%20java.nio.ByteBuffer))
Moreover, i think in this way we could handle the EINVAL case as stated in #232

jneira:
Closed with #744

PR #744, opened 2018-05-05
Fix hsiconv

jneira:
* Fix errors in recoding when overflow case is triggered (#740)


rahulmutt:
Thanks for the fix @jneira! Have you noticed any performance impact from switching from a bulk buffer to a single character `CharBuffer` to hold intermediate encoding state?

jneira:
@rahulmutt  ok, i'll apply those changes

jneira:
@rahulmutt i think with 6fa0700 the migration is done

PR #751, opened 2018-06-01
Fix `CanceledKeyException` caused by inappropriate handling of `SelectionKey` in `checkForReadyIO()`

ouromoros:
The code in `checkForReadyIO` has some problems due to Java NIO's quirk behavior. In Java NIO after a `SelectionKey` is canceled, it isn't really cancelled until the next call to `select()`. And at this time if we try to register the related `Channel` with the `Selector` again, it would throw a `CanceledKeyException`.

Our approach would be to make sure an extra `selectNow()` is called after all the keys are iterated over and "half-canceled". This would involve a while loop to take care of the `Channel`s that are unavoidably selected again, and end when at last `selectNow()` returns 0. Then we can say that every `SelectionKey` is appropriately taken care of, and we can register the `Channel`s without exception as we like.

I'm a bit worried about a race condition though. Say some threads that are woken up in the process and try to register its `Channel` **before** we have the chance to call `selectNow()` and cancel the key appropriately.

If this indeed happens, maybe we need to store all the `TSO`s in a `List` and only try to wake them up after the select loop is finished. Or does there exist a better approach? I'm asking for your opinion here.

### DIFF
--- a/rts/src/main/java/eta/runtime/concurrent/Concurrent.java
+++ b/rts/src/main/java/eta/runtime/concurrent/Concurrent.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Iterator;
+import java.util.ArrayList;
 import java.util.concurrent.Future;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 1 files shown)


rahulmutt:
I agree with your approach of putting all of them in an intermediate list and waking them up outside the while loop. Good thinking!

ouromoros:
@rahulmutt Thanks. I think that'll be all for now.

rahulmutt:
Great! Merging.

PR #784, opened 2018-06-23
Use stack2nix for Nix expression

puffnfresh:
Downside to this is now there's a 1.8MB generated file.

Upside is it works with much more recent nixpkgs.


rahulmutt:
While this looks a lot more maintainable in the future, the generated file seems a bit huge. Is there any way we can make it smaller? Is it copying all the derivations for the transitive dependency tree into one file?

puffnfresh:
@rahulmutt yes, that's exactly what it's doing.

puffnfresh:
We could have a separate repository with the stack2nix generated code, and pull it in via Nix. That would keep the 1.8MB generated files outside of the Eta repo.

Jyothsnasrinivas:
Hi @puffnfresh ! I've added you to eta-lang Org. Can you create a repository to store the generated file?

puffnfresh:
Thanks @Jyothsnasrinivas! The generated stack2nix file is now here:

https://github.com/eta-lang/eta-nix

And I force pushed a change to this PR which uses it.

PR #821, opened 2018-07-23
Add missing entries to `cabal.project` file

hvr:

PR #822, opened 2018-07-23
Add project file for cabal

hvr:
See also #821


hvr:
As opposed to #821, I didn't run into any build failures with the `master` branch

PR #823, opened 2018-07-24
 Support for cabal new-build in a nix-shell

hamishmack:
I find this is a sort of "best of both worlds" way of working.

I set it up as `nix-shell -A shells.ghc` because this is what the reflex-platform uses [here](https://github.com/reflex-frp/reflex-platform/blob/develop/docs/project-development.md).  Reflex platform also has `nix-shell -A shells.ghcjs` for compiling with ghcjs.

I've added [nix support to Leksah](https://github.com/leksah/leksah#leksah-an-integrated-development-environment-for-haskell) that uses `shells.ghc` and `shells.ghcjs`.  I am keen to add `shells.eta` too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ release.log
 binaries/
 .gradle
 testing
+/dist-newstyle/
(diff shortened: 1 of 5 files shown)


Jyothsnasrinivas:
@puffnfresh Thanks for reviewing!

PR #833, opened 2018-07-31
Use `:quit` instead of `:exit` in the REPL

yyadavalli:
Will make it consistent with ghci and easier to quit with `:q`

Only a cosmetic change.


Jyothsnasrinivas:
@yyadavalli Thanks for taking the time for this PR.

Here are the definitions of `exit` and `quit`

`exit`: Go out of or leave a place.
`quit`:  Leave (a place), usually permanently.

As you can see that `exit` makes more sense in this regard as developers are more likely to come back to the REPL after leaving temporarily.

Now I understand that your muscle memory may be adjusted to `:quit`. If so, there is an option to change the defaults by creating the file `~/.eta_repl` with the following contents:
```
def quit const (return ":exit")
```

PR #852, opened 2018-08-19
change FiberYieldException code in capability

ouromoros:
This change in the handling of `FiberYieldException` is needed for implementing exception handling in Fiber.
This will break the previous implementation of `yiledFiber`, so `eta-fibers-dev` would also need to be updated.


rahulmutt:
Looks good, thanks @ouromoros!

Issue #892, opened 2018-10-25
Add new case for foreign exports: allow generic type parameter as result 


jneira:
## Description

I've tried to define a foreign export with this shape:

```haskell
foreign export java "@static org.dhall.eta.Input.withType" jinput
  :: (ja <: Object) => JInputType ja -> JString -> IO ja
jinput jty = undefined
``` 
(same behaviour replacing `IO ja` with `Java x ja`, being x a free variable or a bound one) 
## Expected Behavior
The compiler should generate a java method like:
```java
public class Input {
  public static <A> A withType(InputType<A> ty, String s) {
   ..... 
  }
}
```

## Actual Behavior
Eta throws a compile error:
```
Building library for dhall-eta-0.1.0..
 Compiling Dhall.Eta
 src\main\eta\Dhall\Eta.hs:45:1:  error:
    Unacceptable result type in foreign declaration:
      ‘ja’ is not a data type
    When checking declaration:
      foreign export java "@static org.dhall.eta.Input.withType" jinput
        :: (ja <: Object) => JInputType ja -> JString -> IO ja
 foreign export java "@static org.dhall.eta.withType" jinput
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Possible Fix
Add the case to foreign exports

## Context
Implementing exports to wrap `dhall-haskell` in https://github.com/eta-lang/dhall-eta/blob/java-types/src/main/eta/Dhall/Eta.hs#L45-L47

## Your Environment

* Did you install an older version of Eta/Etlas before? 
  * Yes
* Current Eta & Etlas version:
  * Both built by myself from master versions
* Operating System and version:
  * Windows 10
* Link to your project:
  * https://github.com/eta-lang/dhall-eta/tree/java-types

jneira:
As a workaround i have to wrap the type paremter in a dummy wrapper:
```haskell
data JIdentity a = JIdentity (@org.haskell.types.Identity a)
   deriving (Class,Show,Eq)

foreign import java unsafe "@new" newIdentity  :: (a <: Object)
                                               => a -> JIdentity a

foreign import java unsafe "get" jrunIdentity :: (a <: Object)
                                              => JIdentity a -> a

instance Class a => JavaConverter (Identity a) (JIdentity a) where
  toJava = newIdentity . runIdentity
  fromJava = pure . jrunIdentity

[...]

PR #893, opened 2018-10-29
Allow type variables in ffi export result type

jneira:
## Description
I assimilate check for result type between imports and exports
## How Has This Been Tested?
Compiling a project with the ffi export to be allowed: https://github.com/jneira/eta-test/blob/print-ffi/src/Foreign.hs 

## Types of changes
- [X] Bug fix (non-breaking change which fixes partially the issue #892)

## Checklist:
- [x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/compiler/Eta/DeSugar/DsForeign.hs
+++ b/compiler/Eta/DeSugar/DsForeign.hs
@@ -645,6 +645,10 @@ dsFExport closureId inheritsFamTyCon famInstEnvs co externalName classSpec mod =
                                           []
                                           (ret closureType))))
           closureId
+      varBoundsArgs =
+        mconcat $ replicate (length tyVarBounds)
+                $ invokestatic (mkMethodRef trivialExtendsClass trivialExtendsMethod
+                                [] (ret closureType))
       boxedArgs =
         if length argFts > 5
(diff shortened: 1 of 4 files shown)


rahulmutt:
I'll go ahead and merge this PR. Please do add some tests to the test suite that verify that the corresponding issues are fixed - thanks!

jneira:
>  I think the reason this trivial extends dictionary trick is working is that the extends dictionary is aggressively inlined at the callsite so the dictionary itself is pretty much never used

Mmm but the default implementation of `Extends` at runtime is effectively the id no?, it returns the same object but as another type. So once the typechecker gives license to do the cast at compile time, i think is safe to use id and simply pass the object around.
The problem is it depends on the *default implementation* of Extends so if user set another impl for an instance, f.e
```haskell
instance Extends BigInteger JInteger where
  superCast bi = toInteger bi
  unsafeCast i = newBigInteger i
```

rahulmutt:
It's not effectively `id` because unwrapping and rewrapping happens with different data constructors - that's why unconditional inlining is the key to making the hack work (for now). Once the new implementation for `Extends` is ready, we can have a much cleaner solution.

